PR #19, opened 2022-12-13
Update Client: Align with server and centralize data handling

martin-fleck-at:
- Provide app state to render something while we request the data
- Align data from client with data from server in single file
- Use React context to centralize client and data handling
- Remove most data states within components
- Use React useMemo to derive data from central state
- Use React memo on component to avoid re-render on same props
- Update tests to re-render as we now do not have state changes


martin-fleck-at:
@ivy-lli This is the code that we can go through tomorrow. If you have some time in advance, it would be great if you could have a look because if everything is alright with you then I'd start with splitting the single package into multiple packages based on this commit.

ivy-lli:
> @ivy-lli This is the code that we can go through tomorrow. If you have some time in advance, it would be great if you could have a look because if everything is alright with you then I'd start with splitting the single package into multiple packages based on this commit.

@martin-fleck-at I'm currently at reviewing your changes. As far as I get I looks good :+1: 
It would be great if you can explain a little bit more tomorrow. But I think you can proceed for now.

ivy-lli (approved):
Looks good. I will merge it.

PR #20, opened 2022-12-13
Create monorepo with several packages to ease integration work

martin-fleck-at:
- Split single package into several packages managed by lerna
-- core: contains data
-- client: contains clients for the two servers
-- react-ui: contains the React app, components and client/data context
-- standalone: contains the browser standalone page


martin-fleck-at:
@ivy-lli This change probably breaks the CI, maybe we can have a quick discussion about it tomorrow.

client/packages/react-ui/src/components/InscriptionEditor.tsx
```diff
@@ -1,15 +1,15 @@
...
+import editorIcon from '../../style/icons/user-dialog.svg';
+import '../../style/InscriptionEditor.css';
 import { EditorProps } from './props/editor';
 
 const Header = (props: EditorProps) => (
   <>
     <div className='header'>
       <div className='header-title'>
-        <div className='header-editor'>{props.title}</div>
+        <div className='header-editor'>rf{props.title}</div>
```
ivy-lli, line 12: ?

client/packages/react-ui/src/types/svg.d.ts
```diff
@@ -0,0 +1,10 @@
+declare module '*.svg' {
```
ivy-lli, line 1: For what is this?

ivy-lli (commented):
Yes I think it would be nice if we discuss this in general this afternoon.

But that I don't forget it, some questions or remarks:
- Why are the index exports not used? The imports from other packages are always (or sometimes direct to the generated lib files)
- Why did you move the specific component css files away from the components? Is this not more complicated?
- I think the CI build is failing because you need to run 'yarn all' in the jenkins pipeline
- I general wonder a bit why there are so many different packages (core, react-ui, client)
- I would rename the 'examples' folder to 'integrations'

client/packages/react-ui/src/components/InscriptionEditor.tsx
martin-fleck-at: Sorry, this was committed by mistake. I did a 'watch and update' test and seems like this sneaked into the commit. Sorry for that!

client/packages/react-ui/src/types/svg.d.ts
martin-fleck-at: We need this module declaration to support import statements for '.svg' files.
This may sometimes also be provided by a build system, e.g., see type declarations in CRA: https://github.com/facebook/create-react-app/blob/main/packages/react-scripts/lib/react-app.d.ts#L47 

I will add a comment to the file to clarify this.

client/packages/react-ui/src/components/InscriptionEditor.tsx
ivy-lli: No problem 😀

martin-fleck-at:
@ivy-lli Thank you for the quick review! Let's discuss the individual points in the meeting today, I think most can be handled quickly anyway ;-)

ivy-lli:
I just played a bit with this change, maybe we can also discuss the following points in the meeting:

- I was no longer able to set breakpoints in VSCode except in the standalone package
- HMR is no longer working, sad but maybe this is no longer possible? A watch mode is also ok for me.

ivy-lli:
I just figured out how to fix the breakpoint problem, just remove the `/src` from the launch.json:
https://github.com/axonivy/next-gen-inscription/blob/split-packages/client/.vscode/launch.json#L12

ivy-lli:
I assume this has something to do with false configuration. Currently, jest is still triggered by the react-script package. I assume that this can not handle monorepos or maybe only can not handle this monorepo structure here. 

I think the only way to fix this properly is to get rid of the react-script stuff...
But I think this is something for another PR.

PR #37, opened 2023-01-06
XIVY-10351 Support all inscription for each element

ivy-lli:
- Add Icon package
- Improve styles
- Add some common placeholder tabs
- VsCode Jest extension: autorun only for test file save


ivy-cst (approved):
Looks very nice :+1:

PR #43, opened 2023-01-11
 XIVY-10351 Create TaskTab

ivy-lli:
- Add checkbox widget (persist option, code tab: disable persmission)
- Add select widget
- Add ability to resize CodeEditor widget (code tab)
- Refactor table parts (and add SelectCell)


client/packages/core/src/data/task-data.ts
```diff
@@ -0,0 +1,56 @@
...
+    persist: boolean;
+    task: {
+      name: string;
+      description: string;
+      category: string;
+      priority: string;
+      expiry: {
+        timeout: string;
+        error: string;
+        responsible: {
```
ivy-rew, line 32: I think we should align here: as @ivy-lmu is soon implementing this in the backed: 
I thought that another structure may be easier:

```
responsible: {
  mode = "role/user/..." <-- enum
  assignee = "Everybody/System/... " or "in.myUser" ...
}
```
What do you think about it?

ivy-lli: Yes this is the current state from the backend.
And yes something you proposed should be the future :+1:

client/packages/editor/src/components/tabs/CodeTab.tsx
```diff
@@ -2,18 +2,28 @@ import { useData } from '../../context';
 import { TabProps, TabState } from '../props';
 import LabelInput from '../widgets/label/LabelInput';
 import { CodeEditor } from '../widgets';
+import Checkbox from '../widgets/checkbox/Checkbox';
+
+const SECURITY_SYSTEM = 'system';
```
ivy-rew, line 7: I have a task for that: hopefully the client soon can forget this special value :relaxed: 

i think we will introduce:
`checkPermissions = true/false`
or if someone has a preference:
`asSystem = true/false`

client/packages/editor/src/components/widgets/select/ResponsibleSelect.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+  return (
+    <div className='responsible-select'>
+      <Select label='Responsible' items={responsibleItems} value={selectedResponsible} onChange={item => setResponsible(item.value)}>
+        <>
+          {selectedResponsible?.value === 'ROLE' && <Select label='Role' items={roleItems} value={selectedRole} onChange={() => {}} />}
+          {selectedResponsible?.value === 'ROLE_FROM_ATTR' && <input className='input' />}
+          {selectedResponsible?.value === 'USER_FROM_ATTR' && <input className='input' />}
+        </>
+      </Select>
```
ivy-rew, line 32: things look so easy if solved with REACT+TS ... this component was much more complex back in the SWT days :+1:

client/packages/editor/src/components/widgets/select/Select.css
```diff
@@ -0,0 +1,49 @@
...
+  gap: 0.3rem;
+  width: 100%;
+}
+.select-button:focus {
+  box-shadow: var(--focus-shadow);
+}
+.select-menu {
+  overflow: auto;
+  width: 100%;
+  max-height: 20rem;
```
ivy-rew, line 26: I'm totally outdated in CSS ... never heard of 'rem' values

client/packages/editor/src/components/widgets/table/DocumentTable.tsx
```diff
@@ -89,31 +88,19 @@ const DocumentTable = (props: { data: Document[]; onChange: (change: Document[])
...
-                </button>
-                <button aria-label='Browse'>
-                  <IvyIcon icon={IvyIcons.Search} />
-                </button>
-                <button aria-label='Open URL'>
-                  <IvyIcon icon={IvyIcons.GoToSource} />
-                </button>
-              </span>
-            </TableCell>
+            <ActionCell
```
ivy-rew, line 91: cool generalization in here ...

client/packages/editor/src/components/widgets/table/header/TableHeader.css
```diff
@@ -0,0 +1,11 @@
+.table .table-column-header {
+  padding: 10px;
+  text-align: start;
+  font-size: 10px;
```
ivy-rew, line 4: maybe we should avoid absolute values in a concrete widget ... making the inscription editor easier to access on any device ... even in SWT the user could choose the font-size for the editors. If we go back to static values, we will loose users with bad eye-sight :)

ivy-rew (approved):
looks good :100:

client/packages/editor/src/components/tabs/CodeTab.tsx
ivy-lli: both are good for me. 
Maybe the one which is more the truth? I mean does this flag mean that there are no permission checks (but the same user or is the script step run with the system user?

client/packages/editor/src/components/widgets/select/Select.css
ivy-lli: root element font size. mostly in browsers 16px but can be changed by users. So if a user uses a bigger font-size also this rem value increases.

client/packages/editor/src/components/widgets/table/header/TableHeader.css
ivy-lli: yes I need to change this px values to rem or to another responsive unit :)

client/packages/core/src/data/task-data.ts
ivy-lli: the first one. It's ok if you use the second one to save, but it's really difficult to check which value is available and what to render for it...

ivy-rew: good, yes I think the first one would be the best. Naturally users will think of the data being structured as it is presented on the UI ... keeping the 'mode' and the 'assignee' separate, is also on the plain JSON easier to use. As it is less easy to setup a corrupt configuration :)

client/packages/editor/src/components/widgets/select/Select.css
ivy-rew: thanks for the update :love_letter:  @ivy-lli

ivy-lmu:
Cool 👍

PR #44, opened 2023-01-12
Add VS Code integration extension and integration example based on GLSP

martin-fleck-at:
- Leverage standalone build process to produce VS Code webview content 
  - Use environment variable to control which content is created
  - Webview content listens to dedicated messages from VS Code ('pid')
  - Webview content sends message to VS Code when ready ('ready') 
  - Add build manifest to list webview JavaScript and style files

- Add Inscription Editor integration extension 
  - Provide several commands and shortcuts to open inscription editor 
  - Create webview panel on open command and fill with content
  - Restrict access to external sources for security reasons
  - Provide custom styling for VS Code integration

- Add GLSP Workflow example to VS Code 
  - Provide webview content in dedicated webpack-controlled package 
  - Create extension package to provide workflow extension with webview

- Integrate GLSP workflow example with Inscription Editor 
  - GLSP Workflow extension returns custom API on 'activate' 
  - Inscription Editor extension uses API to track GLSP selection 
  - On selection change, a 'pid' message is sent to the editor webview


martin-fleck-at:
@ivy-cst This is the VS Code integration. I based it on top the Theia integration because I wanted to re-use the `AutoSaveApp` that I added there in the core package but other than that there are no real dependencies between the two integrations.

ivy-lli:
Thanks @martin-fleck-at 
As this is integrated now in the https://github.com/axonivy/vscode-extensions repo I will close this PR and at least stale the branch.

PR #47, opened 2023-01-13
XIVY-10351 Meta changes

ivy-lli:
- Change meta method of dialog starts
- Add meta for roles


ivy-lli:
@ivy-rew 
Currently I have the role meta request on 'meta/workflow/roles' and I expect the following structure:
```javascript
{
  name: string;
  displayName: string;
}[]
```
But I can of course change that :)

PR #52, opened 2023-01-16
XIVY-10351 Different task options

ivy-lli:


packages/editor/src/components/tabs/common/customfield/CustomFieldTable.tsx
```diff
@@ -2,14 +2,18 @@ import { CustomField, CustomFieldType } from '@axonivy/inscription-protocol';
...
+  EditableCell,
+  SelectCell,
+  Table,
+  TableHeader,
+  TableHeaderSorted,
+  TableCell,
+  ActionCell,
+  TableFooter,
+  TableAddRow
+} from '../../../../components/widgets';
```
ivy-rew, line 16: cool; maybe a specific 'table' namespace would be even better? components/widgets/table ...

ivy-rew, line 16: is there a way in TS cluster similar import: so that we can just consume all 'table' widget features at once?

ivy-rew (approved):

PR #53, opened 2023-01-17
 XIVY-10351 Add multi tasks tab

ivy-lli:
- Add task switch gateway mock data
- Load task data from 'config/task' or 'config/tasks'


packages/editor/src/context/useTaskData.ts
```diff
@@ -0,0 +1,13 @@
+import { TaskDataAccess } from '@axonivy/inscription-protocol';
+import { createContext, useContext } from 'react';
+import { useData } from './useData';
+
+export const TaskDataContextInstance = createContext<string>('config/task');
+
+export function useTaskData<K extends keyof TaskDataAccess>(
+  path: K
+): [TaskDataAccess[K], TaskDataAccess[K], (newData: TaskDataAccess[K]) => void];
+export function useTaskData<T>(path: string): [T, T, (newData: T) => void] {
```
ivy-rew, line 10: typescript can get complex too :sweat_smile:

packages/protocol/src/data/task-data.ts
```diff
@@ -18,22 +18,27 @@ export interface CustomField {
...
+
 export interface TaskDataAccess {
-  'config/task/name': string;
-  'config/task/description': string;
-  'config/task/category': string;
-  'config/task/priority': string;
+  name: string;
+  description: string;
+  category: string;
+  priority: string;
```
ivy-rew, line 30: looks simple :+1: 
... maybe we should/Could already extract the 'expiry' to sub-interface too ... and keep the selector path a single key.

packages/core/src/mock/data/task-data.ts
```diff
@@ -3,22 +3,31 @@ import { CustomField } from '@axonivy/inscription-protocol';
...
+  };
+}
+
+export interface TasksData {
+  config: {
+    tasks: Task[];
+  };
+}
+
+interface Task {
```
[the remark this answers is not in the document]
ivy-lli: Don't get your point... This is only for the mocked data. currently, only the name is set, so I will remove the other stuff...

packages/editor/src/context/useTaskData.ts
ivy-lli: Yes ;)

packages/protocol/src/data/task-data.ts
ivy-lli: currently not possible, needs to be a string with a "/" as a separator. But I'm also not very happy with this...

ivy-lli:
I merge it for now. I want to do some improvements to the data access anyway later.

PR #58, opened 2023-01-18
XIVY-10343 IvyScript keywords & operators

ivy-lli:


ivy-cst (approved):
Nice. I think `AS` is still missing

weissreto (commented):
👍 There are more stuff you can remove from tokenizer on line 72.
Here we could also add additional stuff for time literals

PR #59, opened 2023-01-19
Use immer

ivy-lli:


ivy-cst (approved):
Very nice :+1: ! Go for it :100:

PR #66, opened 2023-01-26
XIVY-10509 Filter tree table

ivy-lli:


ivy-rew (approved):
Looks good, great feature :+1: 
And it feels much better to do this here in TS ... rather than hacking object-models in SWT and its models

PR #77, opened 2023-02-08
 Add more unit tests

ivy-lli:
- Simplify table test util (only assert tbody rows)
- Fix combobox/select label connection
- Add tab tests
  - Assert tab content
  - Assert tab state


ivy-lmu (approved):
looks good and streamlined 👍 
A bit odd for me is that the test code is next to prod code. But this might be best practice, you know better 😄

ivy-lli:
I know, a little bit strange for me too, but I think for small unit tests it's ok.

Currently, there are two common ways for js tests:
- test file besides the file which is tested
- all test files in a "__tests__" folder

For me, the first option makes more sense for direct components (unit tests) and the second for integration tests.

PR #81, opened 2023-02-15
Fix output code tab test

ivy-lli:


ivy-rew (approved):
great: so all major issues revealed by the new-inscription view are gone :thinking:

ivy-lli:
No, still present are:
- task tab -> cannot enable code check (because of deadlock somehow...)
- output tab -> mappings are not sent to view

PR #85, opened 2023-03-08
Remove deepmerge

ivy-lli:
Other stuff: 

- Add outputTab and caseTab to other elements
- Add boundary elements


packages/editor/src/components/tabs/task/TasksTab.test.tsx
```diff
@@ -44,9 +46,14 @@ describe('TaskTab', () => {
...
     assertState('empty', { tasks: [] });
-    //@ts-ignore
     assertState('empty', { tasks: [{ id: 'TaskA' }] });
-    //@ts-ignore
     assertState('configured', { tasks: [{ id: 'TaskA', name: 'task1' }] });
   });
+
+  function addDefaultTaskData(data?: DeepPartial<TaskData>): DeepPartial<TaskData> | undefined {
+    if (data) {
+      data.tasks = data.tasks?.map(task => deepmerge(DEFAULT_TASK, task));
```
ivy-rew, line 55: ok, looks as if we can get rid of this dependency altogether, once we we task-switch gateway serialized with the new stack in the backend. :+1:

ivy-rew (approved):

packages/editor/src/components/tabs/task/TasksTab.test.tsx
ivy-lli: This is unit test code, so it has nothing to do with the backend. But yes I can move the dependency to the devDependecy section at least :)

PR #88, opened 2023-03-15
Default meta data

ivy-lli:


packages/core/package.json
```diff
@@ -15,6 +15,9 @@
     "vscode-languageserver-protocol": "^3.17.2",
     "vscode-ws-jsonrpc": "^2.0.1"
   },
+  "devDependencies": {
+    "deepmerge-ts": "^4.2.2"
```
ivy-rew, line 19: welcome back :smile:

ivy-rew (approved):

packages/core/package.json
ivy-lli: was never gone, and only as a dev dependency (like test dependency in maven)

ivy-rew: ahh see, thanks for clarifying

PR #97, opened 2023-04-11
use json.schema generated models

ivy-rew:
- generate inscription.ts models at build time
- use some of the popular models that were easy to adapt: Task, Case, CustomFields, ...

IMHO it helps a lot to understand the backend data-structure without reverse engineering:
- simpler to maintain
- forces us to align the naming of the models with the backend
- still the adaption of new "backend" versions are easily to control and steered by an explicit version in the schema-uri, so it feels like a versioned file, but without SCM overhead.

Future elements will profit even more of these models: just think of the RestElement with it's multiple complex types. 
We should not design these objects manually.


ivy-lli:
I have the same opinion as @ivy-cst 
We could still have a yarn script that downloads the latest schema and generates types of it. So we can check in a build step if the schema is still compatible.

package.json
```diff
@@ -20,7 +20,8 @@
     "icons": "yarn --cwd packages/icons build",
     "dev": "yarn standalone start:mock",
     "test": "yarn --cwd packages/editor test",
-    "test:ci": "lerna run test:ci"
+    "test:ci": "lerna run test:ci",
+    "postinstall": "yarn --cwd packages/protocol"
```
ivy-lli, line 24: Rename this script to what it is doing, e.g. "schema:generate"

packages/editor/package.json
```diff
@@ -42,10 +42,12 @@
   "types": "src/index",
   "module": "src/index",
   "scripts": {
+    "prepare": "yarn build:generate",
     "clean": "rimraf lib *.tsbuildinfo",
     "build": "tsc -p build.tsconfig.json && yarn build:copyfiles",
     "build:clean": "yarn clean && yarn build",
     "build:copyfiles": "copyfiles -u 1 \"src/**/*.css\" lib",
+    "build:generate": "yarn --cwd ../../packages/protocol",
```
ivy-lli, line 50: This package shouldn't contain anything about the protocol package!

packages/editor/src/components/tabs/task/useTaskData.ts
```diff
@@ -25,15 +25,15 @@ export function useTaskData(): {
...
-          draft.description = description;
+        ),
+        [setTask]
+      );
+    
+      const updateDescription = useCallback<Consumer<string>>(
+        description =>
+          setTask(
+            produce(draft => {
+              draft.description = description;
```
ivy-lli, line 36: install the prettier extension in vscode, then this wouldn't happen.

packages/protocol/package.json
```diff
@@ -10,11 +10,17 @@
   "types": "src/index",
   "module": "src/index",
   "scripts": {
-    "clean": "rimraf lib tsconfig.tsbuildinfo",
+    "prepare": "yarn generate",
+    "clean": "rimraf lib tsconfig.tsbuildinfo && yarn generate clean",
```
ivy-lli, line 14: why use a script, which not works on windows, when you already have the solution?
`clean: "rimraf lib tsconfig.tsbuildinfo inscription.json src/data/inscription.ts"`

packages/protocol/package.json
```diff
@@ -10,11 +10,17 @@
...
     "build": "tsc",
     "build:clean": "yarn clean && yarn build",
     "type": "tsc --noEmit",
     "lint": "eslint --ext .ts,.tsx ./src",
-    "lint:fix": "eslint --fix --ext .ts,.tsx ./src"
+    "lint:fix": "eslint --fix --ext .ts,.tsx ./src",
+    "generate": "./schemaCodegen.sh"
```
ivy-lli, line 20: I'm still not fan of this script, we should use a node package / js script to download this schema. Windows users can not run bash scripts.

ivy-lli (changes requested):

packages/protocol/package.json
ivy-rew: the script has some minor benefits:
- we do not have to duplicate the path to the generated inscriptions.ts
- we can easily override the generation temporary, by overriding the URI param (without modifying any sources)
-

packages/protocol/package.json
ivy-rew: I can try that, but at least our Windows users can easily run 'curl' or 'rm' . 
To mit it looks nicer to have a single central place to deal with the schema generation

packages/editor/package.json
ivy-rew: gona try to remove it again ... in the intermediate state the builds couldn't work without it. as the this package uses the protocol package. and the inscription.ts sources were not accessible if this build did not run. ... maybe anyway just an intermediate problem, if we're putting the generated code into version control

packages/protocol/package.json
ivy-lli: we still have, if we write this bash code in js, but then everybody can run this script.

package.json
ivy-rew: it's name like this to profit from lifecycle ... if you wan't an explicit name, we'll need to duplicate it:
```
postinstall: generate
generate: yarn --cwd packages/protocol
```

ivy-lli: we have already a lifecycle hook: prepare

there should be a gernarte script and in the prepare statemant this generate hook should be triggered

packages/editor/package.json
ivy-lli: yes but the generation should be triggered in general and not from this package

packages/protocol/package.json
ivy-rew: -> task for a future PR

ivy-lli (approved):
:+1:

PR #98, opened 2023-04-11
remove 'other' editor

ivy-rew:


packages/protocol/src/data/inscription-type.ts
```diff
@@ -21,15 +21,3 @@ export namespace ActivityEditorType {
   export type Bpmn = 'GenericBpmnElement' | 'UserBpmnElement' | 'ManualBpmnElement' | 'ScriptBpmnElement' | 'ServiceBpmnElement' | 'RuleBpmnElement' | 'SendBpmnElement' | 'ReceiveBpmnElement';
   export type All = General | Interface | Bpmn;
 }
-
-export type OtherEditorType = 'ProcessAnnotation' | 'Unknown';
-
-export type InscriptionEditorType = EventEditorType.All | GatewayEditorType | ActivityEditorType.All | OtherEditorType;
-
-export interface InscriptionTypeBeta {
```
ivy-rew, line 29: the idea of this PR is mainly, to get rid of this re-declaration : as the inscriptionType with its valid element-Types already exists out of the generated inscription.ts protocol

packages/editor/src/context/useEditorContext.ts
```diff
@@ -1,17 +1,17 @@
...
-  type: InscriptionTypeBeta;
+  type: InscriptionType;
 }
 
 export const DEFAULT_EDITOR_CONTEXT: EditorContext = {
   pid: '',
   readonly: false,
   type: {
-    id: 'Unknown',
+    id: 'Script',
```
ivy-lli, line 14: Why this change? The default editor context type shouldn't be script right?

ivy-lli: Ah this doesn't exist in the schema right? :see_no_evil:

integrations/standalone/src/mock.tsx
```diff
@@ -5,12 +5,12 @@ import * as monaco from 'monaco-editor/esm/vs/editor/editor.api';
...
 import ReactDOM from 'react-dom/client';
 import { URLParams } from './url-helper';
-import { InscriptionEditorType } from '@axonivy/inscription-protocol';
+import { InscriptionType } from '@axonivy/inscription-protocol';
 
 export async function start(): Promise<void> {
   const theme = URLParams.themeMode();
   const readonly = URLParams.parameter('readonly') ? true : false;
-  const type = (URLParams.parameter('type') as InscriptionEditorType) ?? undefined;
+  const type = (URLParams.parameter('type') as InscriptionType["id"]) ?? undefined;
```
ivy-lli, line 13: don't know what this "id" is, but we normally use single quotes in js. Also something which prettier should fix I think?

packages/editor/src/components/editors/others/other-editors.tsx
```diff
@@ -1,8 +0,0 @@
-import { IvyIcons } from '@axonivy/editor-icons';
-import { OtherEditorType } from '@axonivy/inscription-protocol';
-import { ReactNode } from 'react';
-import NameEditor from '../NameEditor';
-
-export const otherEditors = new Map<OtherEditorType, ReactNode>([
-  ['ProcessAnnotation', <NameEditor icon={IvyIcons.Note} hideTags={true} />]
```
ivy-lli, line 7: now we lose the ability to configure process annotations?

integrations/standalone/src/mock.tsx
ivy-rew: this is the enumeration that lists all available 'element-types'. VS-Code helped me to write that 'id' quote, gonna try again with prettier

ivy-lli (approved):

PR #101, opened 2023-04-12
node based schemaCodegen script

ivy-rew:


packages/protocol/schemaCodegen.js
```diff
@@ -0,0 +1,54 @@
...
+    file.on('finish', () => {
+      file.close();
+      console.log('Downloaded ' + schema);
+    });
+  });
+}
+
+function codegen() {
+  const { exec } = require('child_process');
+  exec('yarn json2ts ' + schemaFile + ' ' + tsOut, (error, stdout, stderr) => {
```
ivy-lli, line 36: I think you can import this directly (with require or import) and directly call it, instead of execute yarn

packages/protocol/schemaCodegen.js
```diff
@@ -0,0 +1,54 @@
...
+      return;
+    }
+    console.log(`${stdout}`);
+    patchNull();
+  });
+}
+
+function patchNull() {
+  const contents = fs.readFileSync(tsOut);
+  const search = new RegExp('\\?:', 'g');
```
ivy-lli, line 48: /\\?:/g

ivy-lli (approved):
cool

PR #102, opened 2023-04-13
refactor: use promise and reduce IO

ivy-rew:


packages/protocol/schemaCodegen.js
```diff
@@ -16,29 +15,35 @@ if (args.length > 0) {
...
-    response.pipe(file);
-    file.on('finish', () => {
-      file.close();
-      console.log('Downloaded ' + schema);
-      codegen();
-    });
-  });
+  console.log(`loading ${uri}`);
+  const download = new Promise(result =>
+    http.get(uri, res => {
```
ivy-lli, line 25: maybe you can use also the "fetch" method here? but I'm not sure :)

ivy-lli (approved):

packages/protocol/schemaCodegen.js
ivy-rew: unfortunately not, with node 17 as preview, and starting from node 18 it would work. but we are on 16.20

ivy-lli: Ahh ok :see_no_evil:, wasn't sure if its already supported in node ;)

PR #105, opened 2023-04-14
generate meta types too

ivy-rew:
makes use of the schema for meta types.... should get green once the schema is published from the core :)


ivy-lli (approved):
looks good

PR #136, opened 2023-05-15
XIVY-11279 Add undo function

ivy-lli:
- Add accordion tests
- Add undo function (only Name section at the moment)
inscription-view-undo


packages/editor/src/components/tabs/name/NameTab.tsx
```diff
@@ -13,24 +13,45 @@ function useNameTabValidation(): InscriptionValidation[] {
...
+        reset={{ data: data.description, initData: initData.description, resetData: () => updateDescription(initData.description) }}
+        message={descriptionValidation}
+      >
         <Textarea rows={2} id='description' value={data.description} onChange={change => updateDescription(change)} />
       </Fieldset>
-      <Fieldset label='Means / Documents' htmlFor='documents'>
+      <Fieldset
+        label='Means / Documents'
+        htmlFor='documents'
+        reset={{ data: data.docs, initData: initData.docs, resetData: () => updateDocs(initData.docs) }}
```
ivy-rew, line 53: looks to me as if we sort of duplicate:
- the data.binding 
- the bound update method
... which raises the chance to forget things. 

wouldn't it  be cooler, if we could generalize this in the widget; rather than on the glue code? 
looks to me as if the 'widget' could hold it's init data and fire a 'onChange' for resetting the value.

ivy-rew (approved):

packages/editor/src/components/tabs/name/NameTab.tsx
ivy-lli: I know what you mean. But I'm not sure how I can achieve this...



**What is not possible:**
That we apply the data to the fieldset and this one gives it to it's children, because the Fieldset doesn't define its child itself.

**What can be possible:**
That the Fieldset defines a context where the Input (Textarea component) reads from. But I'm not sure if this really simplifies the code, and if this is a the common way to solve it...

ivy-rew: ok for me to keep it as is, it we don't think that the code really gets simpler. 
maybe we can keep it as a "polishing" task for upcoming stories. as we are currently mainly trying to adopt the UI concept.

ivy-lli: What we can do, is that the fieldset and the different widgets in in takes the same data object, e.g like the following:

```typescript
const NameTab = (props: { hideTags?: boolean }) => {
  const { data, initData, updateName, updateDescription, updateDocs, updateTags } = useNameData();
  const [nameValidation, descriptionValidation] = useNameTabValidation();

  const nameData = { data: data.name, initData: initData.name, updateData: updateName };
  const descriptionData = { data: data.description, initData: initData.description, updateData: updateDescription };
  const docsData = { data: data.docs, initData: initData.docs, updateData: updateDocs };

  return (
    <>
      <Fieldset label='Display name' htmlFor='displayName' data={nameData} message={nameValidation}>
        <Textarea rows={1} id='displayName' data={nameData} />
      </Fieldset>
      <Fieldset label='Description' htmlFor='description' data={descriptionData} message={descriptionValidation}>
        <Textarea rows={2} id='description' data={descriptionData} />
      </Fieldset>
      <Fieldset label='Means / Documents' data={docsData}>
        <DocumentTable data={docsData} />
      </Fieldset>
      {!props.hideTags && (
        <CollapsiblePart collapsibleLabel='Tags' defaultOpen={data.tags !== undefined && data.tags.length > 0}>
          <Tags tags={data.tags ?? []} onChange={updateTags} />
        </CollapsiblePart>
      )}
    </>
  );
};
```

PR #145, opened 2023-05-19
XIVY-11447 Add Alternative element

ivy-lli:


ivy-lli:
Missing:
- Use of connector meta api
- Integration test

PR #158, opened 2023-05-25
XIVY-11447 conditions location

ivy-rew:
fixed


ivy-lli (approved):
Ahhh was not sure if this was even needed at all...
But I think the type is misleading here. This is meant for e.g. the java type on the mapping tables.

ivy-rew:
> 

looking at the backend, the "type" isn't interpreted anywhere ... so I don't know what to do with it :)

PR #177, opened 2023-06-15
XIVY-10873 Fix change result map on param change

ivy-lli:


packages/editor/src/components/parts/result/ResultPart.tsx
```diff
@@ -26,8 +26,16 @@ const ResultPart = ({ hideParamDesc }: { hideParamDesc?: boolean }) => {
...
   useEffect(() => {
-    client.resultMapping(editorContext.pid).then(mapping => setMappingInfo(mapping));
-  }, [client, editorContext.pid]);
+    client.resultMapping(editorContext.pid).then(mapping => {
+      const resultType = mapping.variables[0].type;
+      if (mapping.types[resultType]?.length !== data.result.params.length) {
+        mapping.types[resultType] = data.result.params.map<Variable>(param => {
+          return { attribute: param.name, type: param.type, simpleType: param.type, description: param.desc };
+        });
+      }
```
ivy-lli, line 35: this is the fix.
If the size of the types does not match the param size (new param was added or existing one removed), then the params will replace the types of the backend

PR #183, opened 2023-06-21
Email header part

ivy-lli:
Solution from me, currently without any tests


ivy-cst (commented):
Beside the tests also the mock data is missing.
I'm not sure if just Header is ok for the name:
- Seems specific to Email and can not be used somewhere else I think.

packages/protocol/src/data/part-data.ts
```diff
@@ -52,3 +52,7 @@ export interface SignalCatchData {
   signalCode: string;
   attachToBusinessCase: boolean;
 }
+
+export interface HeaderData {
+  headers: MailHeaders;
```
jackra1n, line 57: I did not see that this already existed 😮
otherwise ther rest did almost the same 😃

PR #190, opened 2023-06-28
XIVY-11675 Reinvent CodeEditor widget

ivy-lli:
New widgets
- ScriptInput (single line code editor, active on focus)
- ScriptArea (resizable code editor, always active)
- MacroInput (single line code editor, active on focus)
- MacroArea (Resizable code editor, active on focus)
- ScriptCell (with popover single line code editor)

Other stuff done:
- Link everywhere the label with the input (via useFieldset hook)
- Change test code for new CodeEditor widgets


weissreto:
Cool implementations. 

As an inscription mask programmer, I may not be interested how a script editor is implemented (resizable) or how it behaves (focus). All I know is I need to have a macro input field, script input field, macro text area, script text area. How this is implemented is not important for me and may event change in future. By naming the basic buidling blocks independent of the implementation we can change the implementation later without the need to change all inscription mask once again.

Therefore I would suggest that we introduce these five widgets:
ScriptInput
ScriptArea (ScriptEditor?)
MacroInput
MacroArea (MacroEditor?)
ScriptCellEditor

These are used in the inscription masks. How they are implemented is hidden from the user and may change in future.

integrations/standalone/tests/integration/parts/call.ts
```diff
@@ -17,7 +17,7 @@ export class CallTester implements PartTest {
   }
   async fill(page: Page) {
     await ComboboxUtil.select(page, this.selectValue, this.selectLabel);
-    await TableUtil.fillExpression(page, 2, '"test"');
```
weissreto, line 20: I general we talk about the expression and not the code in mapping tables. 
I find the old naming better.

packages/editor/src/components/widgets/code-editor/index.ts
```diff
@@ -1 +1,4 @@
 export { default as CodeEditor } from './CodeEditor';
+export { default as ResizableCodeEditor } from './ResizableCodeEditor';
+export { default as SingleLineCodeEditor } from './SingleLineCodeEditor';
```
ivy-rew, line 3: We can keep this name and it's very intuitive, but still a small remark.
In the process.model apis and bpm.exec we tried to distinguish single line codes from long statements by calling single liners "expression" (ScriptExpression, MacroExpression) and so on. 
If you like the term, you may adapt it here to, and name it "CodeExpressionEditor"?

integrations/standalone/tests/utils/table-util.ts
```diff
@@ -19,10 +19,16 @@ export namespace TableUtil {
     }
   }
 
-  export async function fillExpression(page: Page, rowIndex: number, value: string) {
-    await page.locator('tbody').getByRole('row').nth(rowIndex).getByRole('textbox').click();
+  export async function fillCode(page: Page, rowIndex: number, value: string) {
```
weissreto, line 22: either code or expression but not both

ivy-rew (approved):
Feels good, now we have concrete widgets almost as we have them in SWT ;-) ... so it feels like going into the same full featured direction :+1:

packages/editor/src/components/widgets/tag/Tags.css
```diff
@@ -36,100 +36,3 @@
 .tag-add:focus-visible {
   box-shadow: var(--focus-shadow);
 }
-
```
weissreto, line 39: Not needed anymore. some dead bodies here?

ivy-rew:
> Cool implementations.
> 
> As an inscription mask programmer, I may not be interested how a script editor is implemented (resizable) or how it behaves (focus). All I know is I need to have a macro input field, script input field, macro text area, script text area. How this is implemented is not important for me and may even change in the future. By naming the basic building blocks independent of the implementation we can change the implementation later without the need to change all inscription mask once again.
> 
> Therefore I would suggest that we introduce these five widgets: ScriptInput ScriptArea (ScriptEditor?) MacroInput MacroArea (MacroEditor?) ScriptCellEditor
> 
> These are used in the inscription masks. How they are implemented is hidden from the user and may change in future.

I can second that the naming seems to technical right now. I like your proposals ... I'd go with the 'Input/area/cellEdit' naming. At least I'd instantly know which one too choose.

weissreto (changes requested):

packages/editor/src/components/widgets/tag/Tags.css
ivy-lli: No, moved to Popover.css

integrations/standalone/tests/integration/parts/call.ts
ivy-lli: ok i change it back 👍

integrations/standalone/tests/utils/table-util.ts
ivy-lli: same as above

ivy-rew (approved):

weissreto (approved):
Cool, thanks for the work

PR #192, opened 2023-06-30
Attribute browser local

ivy-lli:


packages/editor/src/components/widgets/code-editor/ScriptInput.tsx
```diff
@@ -1,17 +1,26 @@
...
+  const { isFocusWithin, focusWithinProps } = useCodeEditorOnFocus();
+  const browser = useBrowser();
+  const { setEditor, modifyEditor } = useModifyEditor();
+
+  return (
+    <div className='script-input' {...focusWithinProps}>
+      {isFocusWithin || browser.open ? (
+        <>
+          <SingleLineCodeEditor {...props} onMountFuncs={[setEditor]} />
+          <Browser {...browser} types={['attribute', 'cms']} accept={modifyEditor} />
```
ivy-rew, line 17: maybe you could/should generalize that part 'types=attr+cms' ... as it is now equally stated on all script editors.

ivy-rew (approved):

packages/editor/src/components/widgets/code-editor/ScriptInput.tsx
ivy-lli: Yes not part of this pr. I will address that as soon as I implement a better UI with real data 👍

PR #200, opened 2023-07-04
XIVY-11675 Add own monaco language for ivy macros

weissreto:
Add own Monaco language for ivy macros and provide highlighting of <% and %>

Also support that %> is auto inserted if <% is inserted

ivyScript and ivyMacro language is served by the same LSP


ivy-lli (approved):
As said below in the comments, otherwise good. Thanks!

PR #203, opened 2023-07-11
XIVY-11675 Small review fixes

ivy-lli:
- Fix suggestion widget position
  - Normal monaco's need to have this `fixedOverflowWidgets` option, but not the one in the mapping tree
- Fix Safari loose focus by clicking on the attribute browser
   Looks like Safari is the next IE... 🤮
   https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#clicking_and_focus
- Fix react warning about missing key attribute in attribute browser
- Fix css gap in collapsable


ivy-rew (approved):

ivy-lmu (approved):
thanks 👍

PR #206, opened 2023-07-12
XIVY-11598 Fix script location

ivy-lli:
- Create context prop for ScriptInput with location and type. (This fixes the condition attribute browser, because it no longer sends the request with the type in the location)


weissreto (approved):
👀✔

PR #208, opened 2023-07-12
Feature/XIVY-11675 page objects

weissreto:


integrations/standalone/tests/integration/parts/task.ts
```diff
@@ -17,28 +14,32 @@ export class TasksTester implements PartTest {
   partName() {
     return 'Tasks';
   }
-  async fill(page: Page) {
+  async fill(part: Part) {
     for (const task of this.tasks) {
-      await TabUtil.change(page, task.tab);
-      await task.test.fill(page);
+      const tab = (part as Accordion).tab(task.tab);
+      await tab.switchTo();
```
ivy-lli, line 20: maybe not possible (I didn't see all the code yet), but I would expect a general tab object and the switchTo method should get the name of the tab. So the reading is the tabs should switch to the tab name...

integrations/standalone/tests/integration/parts/task.ts
```diff
@@ -58,86 +59,113 @@ export class TaskTester implements PartTest {
...
+    await responsible.expectValue('"Teamleader"');
 
-    await SelectUtil.select(page, 'High', 1);
+    await part.select('Priority').expectValue(/High/);
 
-    await CollapseUtil.open(page, 'Options');
-    await page.getByRole('checkbox').check();
-    await page.getByLabel('Delay').fill('delay');
+    const options = part.section('Options');
+    await expect(options.checkbox('Skip Tasklist')).toBeChecked();
```
ivy-lli, line 110: I think this checkbox object should also provide a method to check the state, e.g.:
`await options.checkbox('Skip Tasklist').expectChecked(true)`

integrations/standalone/tests/integration/parts/task.ts
ivy-lli, line 1: Yes! Looks very promising 😎 
Maybe a thought more in this way, and if then I would do that in a later PR:
I think this 'part tests' should also be such page objects. Because if so, we can initialize all this elements in the constructor (e.g. the macroInput('Name')) so we don't need to define this object in each method again. And this would also ease this "many tasks" / "single task" problematic I think.

integrations/standalone/tests/pageobjects/CodeEditor.ts
```diff
@@ -0,0 +1,50 @@
...
+  private readonly label: string | undefined;
+  private readonly parentLocator: Locator;
+  private readonly page: Page;
+  private readonly contentAssist: Locator;
+  private readonly value: Locator;
+
+  constructor(page: Page, parentLocator: Locator, activateOnFocus: boolean, label: string | undefined) {
+    this.page = page;
+    this.parentLocator = parentLocator;
+    this.label = label;
```
ivy-lli: Instead of this, you can also do it like this:
```js
export class CodeEditor {
  private readonly contentAssist: Locator;
  private readonly value: Locator;
  constructor(readonly page: Page, readonly parentLocator: Locator, activateOnFocus: boolean, readonly label?: string) {
    this.contentAssist = parentLocator.locator('div.suggest-widget');
    this.value = activateOnFocus ? this.locator() : this.locator().getByRole('textbox');
  }
```

If you use the readonly param in the constructor it automatically creates a parameter for you and does the assignment.

integrations/standalone/tests/pageobjects/CodeEditor.ts
```diff
@@ -0,0 +1,50 @@
...
+    this.contentAssist = parentLocator.locator('div.suggest-widget');
+    this.value = activateOnFocus ? this.locator() : this.locator().getByRole('textbox');
+  }
+
+  async triggerContentAssist() {
+    await FocusCodeEditorUtil.triggerContentAssist(this.page, this.locator());
```
ivy-lli: I think after every part is using this pageobject we can move the code of this util classes inside here 👍

integrations/standalone/tests/pageobjects/Composite.ts
```diff
@@ -0,0 +1,53 @@
...
+  macroArea(label: string) {
+    return new CodeEditor(this.page, this.locator(), true, label);
+  }
+
+  scriptInput(label: string) {
+    return new CodeEditor(this.page, this.locator(), true, label);
+  }
+
+  scriptArea() {
+    return new CodeEditor(this.page, this.locator(), false, undefined);
```
ivy-lli: if you define the label as `label?: string` instead of `label: string | undefined`, you can remove undefined here.

ivy-lli: Even if this too maybe result in the same thing, they don't mean the same:
- label?: string -> means that this parameter is optional
- label: string | undefined -> means that this parameter needs to be a string or undefined

integrations/standalone/tests/pageobjects/Composite.ts
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  table() {
+    return new Table(this.page, this.locator(), undefined);
```
ivy-lli: same here -> remove undefined by change the parameter to optinal

integrations/standalone/tests/pageobjects/ResponsibleSelect.ts
```diff
@@ -0,0 +1,57 @@
...
+    await this.script.fill(value);
+  }
+
+  async clear() {
+    await this.chooseType('Role');
+    await this.combo.click();
+    await this.locator().getByRole('option', { name: 'Everybody' }).first().click();
+  }
+
+  async chooseType(type: 'Role' | 'Role from Attr.' | 'User from Attr.' | 'Nobody & delete') {
```
ivy-lli, line 26: you can reuse this type from the protocol package (a little bit of typescript magic):
```ts
type ObjectValuesAsUnion<T> = {
  [V in keyof T]: T[V];
}[keyof T];

async chooseType(type: ObjectValuesAsUnion<typeof RESPONSIBLE_TYPE>) {
```

The first one is a typehelper, maybe we should move this to the editor package or something like that...

integrations/standalone/tests/pageobjects/Table.ts
```diff
@@ -0,0 +1,90 @@
...
+  constructor(page: Page, parentLocator: Locator, label: string | undefined) {
+    this.page = page;
+    this.parentLocator = parentLocator;
+    this.label = label;
+    this.rows = this.tableLocator().locator('tbody').getByRole('row');
+  }
+
+  async addRow() {
+    await this.tableLocator().getByRole('row', { name: 'Add row' }).click();
+    return this.row(0);
```
ivy-lli, line 18: This will only work if there was no row before. Maybe there is already one, so we should return the latest row here.

ivy-lli (approved):
Looks definitely like the right way to write such tests.
Thank you very much for your work!

Some small comments, some proposals, some change wishes, some ts wizardry 🧙

PR #210, opened 2023-07-13
XIVY-11485 use new propaged SchemaKeys

ivy-rew:


packages/protocol/src/data/inscription.ts
```diff
@@ -557,3 +558,102 @@ export interface NodeRef {
...
+export interface SchemaKey {
+  OUTPUT: "output";
+  EXCEPTION_HANDLER: "exceptionHandler";
+  CODE: "code";
+  MAP: "map";
+  Alternative: {
+    CONDITIONS: "conditions";
+    [k: string]: unknown;
+  };
+  Cachable: {
```
ivy-rew, line 570: better @ivy-lli ?

PR #211, opened 2023-07-13
XIVY-11521 Add more open actions

ivy-lli:
- Open custom-field config
- Open document

### DIFF
--- a/packages/protocol/src/actions.ts
+++ b/packages/protocol/src/actions.ts
@@ -1,5 +1,9 @@
+export type CustomFieldType = 'START' | 'TASK' | 'CASE';
+
 export interface InscriptionActionHandlers {
   newHtmlDialog: undefined;
   newProcess: undefined;
-  openFile: string;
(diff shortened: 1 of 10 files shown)


packages/protocol/src/actions.ts
```diff
@@ -1,5 +1,7 @@
+export type CustomFieldType = 'START' | 'TASK' | 'CASE';
+
 export interface InscriptionActionHandlers {
   newHtmlDialog: undefined;
   newProcess: undefined;
-  openFile: string;
+  openFile: `pageUri=${string}` | `customField=${CustomFieldType},${string}` | `doc=${string}`;
```
ivy-rew, line 6: this one is also asking for a clear structure as it seems ;-)

ivy-lli: I can provide a bit of type safety here, but this should definitely also be handled by a clear type from the protocol

ivy-rew (approved):

PR #214, opened 2023-07-14
XIVY-11485 Validation location

ivy-lli:
- Use known schema keys
- Add validations to all parts
- Use locations for validations and ivyscript context
- Use page objects for mock playwright tests

### DIFF
--- a/packages/protocol/src/data/inscription-data.ts
+++ b/packages/protocol/src/data/inscription-data.ts
@@ -14,7 +14,8 @@ import {
   StartData,
   TaskData
 } from './part-data';
-import { InscriptionType } from './inscription';
+import { InscriptionType, SchemaKey } from './inscription';
+import { ValuesAsUnionDeep } from '../utils/type-helper';
 
 export type ConfigData = CallData &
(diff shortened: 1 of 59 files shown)


packages/protocol/src/data/inscription-data.ts
```diff
@@ -47,3 +48,9 @@ export interface InscriptionSaveData {
 }
 
 export type ElementType = InscriptionType['id'];
+
+type TaskCaseSchemaKeys = 'name' | 'description' | 'category' | 'expriy';
+
+type EmailSchemaKeys = 'subject' | 'from' | 'to' | 'replyTo' | 'cc' | 'bcc';
```
ivy-rew, line 54: maybe there is another typescript magic way to extract these keys from the properties of the generated inscription objects?

header-properties

ivy-rew (approved):

packages/protocol/src/data/inscription-data.ts
ivy-lli: I try, but will be placed in a follow up PR.

PR #215, opened 2023-07-14
XIVY-11521 consume new action.schema

ivy-rew:


packages/protocol/src/data/inscription.ts
```diff
@@ -497,11 +498,15 @@ export interface InscriptionValidation {
   path: string;
   severity: Severity;
 }
-export interface InscriptionAction {
-  kind: string;
-  payload: string;
+export interface InscriptionActionArgs {
+  actionId: "openEndPage" | "newProcess" | "newHtmlDialog" | "openCustomField" | "openPage";
+  payload: string | OpenCustomField;
   pid: PID;
```
ivy-rew, line 504: can't easily establish a strict correlation from the 'id' to the matching 'payload' ... but still it's getting very comfortable here

packages/protocol/src/actions.ts
ivy-rew, line 1: this is now auto-generated ... and future proof :100:

packages/protocol/src/data/inscription.ts
ivy-lli: I think it is ok if we do that manually, to at least help adopters of these actions. But definitely fine as a start.

ivy-lli (approved):

PR #228, opened 2023-07-21
XIVY-11485 Script types

ivy-lli:
Currently, only the Inscription mask validator knows this.
But our ivyscript lsp currently does know nothing from the input except that evaluated variables.

If we would like to change this, we either need to make the ivyscript lsp more stateful or evaluate the real variable via the location. But if we do the second way, we also need to evaluate the value of a mapping tree, and this is something we definitely know in the frontend.

So I think it does no harm now if we send these types to the backend by now...

Or am I wrong?

### DIFF
--- a/packages/editor/src/components/parts/task/options/TaskListPart.tsx
+++ b/packages/editor/src/components/parts/task/options/TaskListPart.tsx
@@ -1,3 +1,4 @@
+import { IVY_SCRIPT_TYPES } from '@axonivy/inscription-protocol';
 import { Checkbox, ScriptInput, useFieldset } from '../../../widgets';
 import { PathCollapsible, ValidationFieldset } from '../../common';
 import { useTaskData } from '../useTaskData';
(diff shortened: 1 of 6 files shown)


packages/editor/src/components/parts/task/options/TaskListPart.tsx
```diff
@@ -10,7 +11,12 @@ const TaskListPart = () => {
     <PathCollapsible label='Options' defaultOpen={task.skipTasklist} path='delay'>
       <Checkbox label='Skip Tasklist' value={task.skipTasklist} onChange={change => update('skipTasklist', change)} />
       <ValidationFieldset label='Delay' {...delayFieldset.labelProps}>
-        <ScriptInput value={task.delay} onChange={change => update('delay', change)} {...delayFieldset.inputProps} type='Duration' />
+        <ScriptInput
+          value={task.delay}
+          onChange={change => update('delay', change)}
+          {...delayFieldset.inputProps}
+          type={IVY_SCRIPT_TYPES.DURATION}
```
ivy-rew, line 18: we did just the same in the old SWT world. So I don't see this currently as a big problem. So I think it's ok to keep it as is ... and making it easy to change if required later on ... by using these constants :+1: 
duration-swt

ivy-rew (approved):
does no harm; and repeats what we have done with SWT. so it looks good to me; I wouldn't invest any further in this topic right now

PR #230, opened 2023-07-21
XIVY-10544 Add app and pmv to protocol

ivy-lli:


ivy-rew (approved):

integrations/standalone/src/mock.tsx
```diff
@@ -23,7 +23,7 @@ export async function start(): Promise<void> {
     <React.StrictMode>
       <ThemeContextProvider theme={theme}>
         <ClientContextInstance.Provider value={{ client: inscriptionClient }}>
-          <App pid={'1'} />
+          <App app='' pmv='' pid={'1'} />
```
weissreto, line 26: Can we provide these values somehow so that mock works as similar as standalone?
What does {'1'} mean here?

integrations/standalone/src/mock/inscription-client-mock.ts
```diff
@@ -36,7 +35,7 @@ export class InscriptionClientMock implements InscriptionClient {
     return Promise.resolve(true);
   }
 
-  data(pid: PID): Promise<InscriptionData> {
+  data(args: InscriptionContext): Promise<InscriptionData> {
```
weissreto, line 38: args -> context

integrations/standalone/src/url-helper.ts
```diff
@@ -6,6 +6,15 @@ export namespace URLParams {
     return param !== null ? param : undefined;
   }
 
+  export function app(): string {
+    const serverUrl = server();
+    return parameter('app') ?? serverUrl.substring(serverUrl.lastIndexOf('/') + 1);
```
weissreto, line 11: use case of the serverUrl last part?

packages/core/src/inscription-client-jsonrpc.ts
```diff
@@ -38,56 +37,56 @@ export class InscriptionClientJsonRpc extends BaseRcpClient implements Inscripti
     return this.sendRequest('initialize', undefined);
   }
 
-  data(pid: PID): Promise<InscriptionData> {
-    return this.sendRequest('data', { pid });
+  data(args: InscriptionContext): Promise<InscriptionData> {
```
weissreto, line 40: args -> context

packages/editor/src/App.tsx
```diff
@@ -67,7 +62,7 @@ function App(props: AppProps) {
       <div ref={editorRef} className='editor-root' data-theme={theme}>
         <EditorContextInstance.Provider
           value={{
-            pid: props.pid,
+            context: appState.initialData.context,
```
weissreto, line 65: Why has this changed from props to appState?

packages/protocol/src/inscription-client.ts
```diff
@@ -23,24 +22,24 @@ export interface Disposable {
 
 export interface InscriptionClient {
   initialize(): Promise<boolean>;
-  data(pid: PID): Promise<InscriptionData>;
+  data(args: InscriptionContext): Promise<InscriptionData>;
```
weissreto, line 25: args -> context

weissreto (approved):
Overall good change. Some comments and questions

integrations/standalone/src/mock.tsx
ivy-lli: don't know what should be the benefit of this...
The mock does nothing with this information, it is only here to quickly test and develop some ui stuff.

packages/editor/src/App.tsx
ivy-lli: because up to now it is still supported that you request a inscription view only with a pid (this is very helpful for testing as I don't want always provide also the app and the pmv).
Then the resolved app and pmv for the given pid is delivered by the server.

integrations/standalone/src/url-helper.ts
ivy-lli: today you can already change the server url (e.g. if the designer is started on 8082). And in this url the app is normally already included e.g. `localhost:8082/designer`

integrations/standalone/src/mock.tsx
weissreto: My idea was that if we provide here some dummy information we can check if they are sent correctly to the mock server. 
Something like app='testApp' pmv="test$12'

integrations/standalone/src/url-helper.ts
weissreto: If you like to support that please consider URL decoding for the app name

ivy-lli: Hm... honestly I don't think that it is really necessary. I thought it would be a nice idea, but I think you either use only a pid or then a pid with app and pmv...

integrations/standalone/src/mock.tsx
ivy-lli: Hmm... yes this is maybe an idea. But I think this has nothing to do with the mock server. Because we don't need to test the mock server. Or better we cannot test such stuff in the mock server, because this info is never displayed anywhere on the ui. And the mock server is only in charge in the mock webtests...

But a general App.tsx or a custom hock unit test would be good for this. Also to test the error view (e.g. if an unknown type is delivered).

But I will check that in a later PR.

PR #232, opened 2023-07-25
XIVY-10544 Move location and type to ivyscript path

ivy-lli:


packages/editor/src/components/widgets/code-editor/CodeEditor.tsx
```diff
@@ -34,7 +34,7 @@ const CodeEditor = ({ value, onChange, context, macro, onMountFuncs, options, ..
         defaultValue={value}
         value={value}
         defaultLanguage={language}
-        defaultPath={`${language}/${contextPath}?location=${context.location}${context.type ? `&type=${context.type}` : ''}`}
+        defaultPath={`${language}/${contextPath}/${context.location}/${context.type ? context.type : ''}`}
```
weissreto, line 37: Only add / if we have a type as you do it in your test in the other PR?

weissreto (approved):

packages/editor/src/components/widgets/code-editor/CodeEditor.tsx
ivy-lli: Makes no difference in the backend. But I've added a test for this on the engine

PR #235, opened 2023-07-26
Use new page obejcts in the condtion part test

weissreto:
Imrpove also the task part test with a dedicated Task page object


ivy-lli (approved):
👍

ivy-lli (approved):

PR #236, opened 2023-07-26
Add useMeta hook

ivy-lli:
Simplify inscription client api, by reduce all meta methods to one with a custom useMeta hook.

### DIFF
--- a/integrations/standalone/src/mock/inscription-client-mock.ts
+++ b/integrations/standalone/src/mock/inscription-client-mock.ts
@@ -1,26 +1,24 @@
 import {
   DEFAULT_DATA,
-  CallableStart,
-  ErrorMeta,
   InscriptionClient,
   InscriptionData,
   InscriptionSaveData,
   InscriptionType,
   InscriptionValidation,
-  VariableInfo,
-  RoleMeta,
   ElementType,
   ElementData,
-  ConnectorRef,
   InscriptionActionArgs,
-  EventCodeMeta,
-  InscriptionContext
+  InscriptionContext,
+  InscriptionMetaRequestTypes,
+  ScriptingDataArgs,
+  Variable
 } from '@axonivy/inscription-protocol';
 import { Emitter } from 'vscode-jsonrpc';
 import { deepmerge } from 'deepmerge-ts';
 import { DataMock } from './data-mock';
 import { ValidationMock } from './validation-mock';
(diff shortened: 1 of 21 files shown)


ivy-lli:
How said this isn't typed?
It gives you full autocompletion on the path and on the args, returns the correct type for the chosen path, and will also fail if something changes.

But the difference here is that if we introduce a new meta api to the protocol, we only need to do this: Add this new path with the args and return type to the protocol and you get access to it over the useMeta hook.
If we don't have this we need to duplicate this method on every inscription client implementation (the jsonrcp, the mock, and the test), so it will work...

This is not Java this is Typescript. And Typescript has JavaScript under the hood, so we need to accept that some parts work differently in this language...

And by the way, this is only the first part of the React-Query integration. So I need to do this change anyway if we want to use React-Query and don't want to duplicate even more code.

weissreto:
> How said this isn't typed? It gives you full autocompletion on the path and on the args, returns the correct type for the chosen path, and will also fail if something changes.
> 
> But the difference here is that if we introduce a new meta api to the protocol, we only need to do this: Add this new path with the args and return type to the protocol and you get access to it over the useMeta hook. If we don't have this we need to duplicate this method on every inscription client implementation (the jsonrcp, the mock, and the test), so it will work...
> 
> This is not Java this is Typescript. And Typescript has JavaScript under the hood, so we need to accept that some parts work differently in this language...

> 
> And by the way, this is only the first part of the React-Query integration. So I need to do this change anyway if we want to use React-Query and don't want to duplicate even more code.

I'm not saying this is bad. I'd like to understand this change first. 

1.) It is not Java it is Typescript: Yes, but this is only one side of the medal. The interface needs to be implemented on Java side right? Can you link me to the PR on that side?

2.) We need to do this change anyway. I don't think we need to use React-Query anyway if we don't like it. We already implemented it without it. What are the main advantages of using it. 

I think this is major decision and we should agree on this before we go ahead. What are the opinions of the others (cst?).

ivy-lli:
Hm? No this has nothing to do with the backend.
This is only a change in how the client can make meta requests. There has been nothing changes on the backend.

React-Query gives comes with many features like cache data etc, which can improve the user experience a lot. https://tanstack.com/query/latest/docs/react/overview

weissreto (approved):
Thanks for explaining me this change. 
Looks good.

PR #237, opened 2023-07-26
XIVY-10544 Fix pmv param decode

ivy-lli:
Param pmv=workflow-demos%25241 from the Designer integration was not correclty decodet to 'workflow-demos$1'


weissreto (approved):
JScript 😢

PR #239, opened 2023-07-27
Don't show messages global with path

ivy-lli:
With this change messages at the editor top will only be shown if they have no path.

- No longer an up and down while writing something in an input and the message changes (was while the demo)
- Should be a hint that a message from the backend has no path set
- IMO those are no longer needed, because if they have a path they should be shown on the 
  1. accordion
  2. widget (input, etc) 

I also discussed this yesterday with pascal.
What do you think @weissreto?

Before:
Screen Recording 2023-07-27 at 09 39 50

After:
Screen Recording 2023-07-27 at 09 41 19


ivy-lli:
What do you mean with read only mode? 
Do you mean when the project is packed (iar)?

I think it makes no difference because also in readonly mode the widget will be shown as they are with a message if they have one...
What we may can think about is, if we want do display the messages on the summary content if we implement those.

weissreto:
No, I mean if you open the inscription view and you are in the reader mode and not the edit mode. Then the errors should be visible somewhere. I think this is a task for Anais to think about where to display them in this mode.

ivy-lli:
Ah, when the accordions are closed?
Yes of course, as I said, I think they should be displayed somewhere in the summary part...
But currently, at least the accordion shows with the state if it has an error in it.

PR #241, opened 2023-07-28
XIVY-12024 Add macro card text

ivy-lli:
- Show output with card badges for macro expressions if macro editors are in read mode
- Init MacroArea monaco editor with the same height as the read mode was
- Specific CodeEditor page objects

image

### DIFF
--- a/integrations/standalone/tests/pageobjects/CodeEditor.ts
+++ b/integrations/standalone/tests/pageobjects/CodeEditor.ts
@@ -1,23 +1,11 @@
 import { Locator, Page, expect } from '@playwright/test';
 import { FocusCodeEditorUtil } from '../utils/code-editor-util';
 
-export class CodeEditor {
-  private readonly page: Page;
-  private readonly locator: Locator;
-  private readonly contentAssist: Locator;
-  private readonly value: Locator;
+class CodeEditor {
+  protected contentAssist: Locator;
 
-  constructor(page: Page, parentLocator: Locator, activateOnFocus: boolean, label?: string) {
-    this.page = page;
+  constructor(readonly page: Page, readonly locator: Locator, readonly value: Locator, parentLocator: Locator) {
     this.contentAssist = parentLocator.locator('div.suggest-widget');
-    if (label) {
-      this.locator = parentLocator.getByLabel(label, { exact: true }).first();
-    } else if (activateOnFocus) {
-      this.locator = parentLocator.getByRole('textbox');
-    } else {
-      this.locator = parentLocator.getByRole('code').nth(0);
-    }
-    this.value = activateOnFocus ? this.locator : this.locator.getByRole('textbox');
   }
 
   async triggerContentAssist() {
@@ -42,5 +30,33 @@ export class CodeEditor {
 
   async expectContentAssistContains(contentAssist: string) {
     await expect(this.contentAssist).toContainText(contentAssist);
+  }
+}
+
+export class ScriptArea extends CodeEditor {
+  constructor(page: Page, parentLocator: Locator) {
+    const locator = parentLocator.getByRole('code').nth(0);
+    super(page, locator, locator.getByRole('textbox'), parentLocator);
+  }
+}
+
+export class ScriptInput extends CodeEditor {
+  constructor(page: Page, parentLocator: Locator, label?: string) {
+    let locator = parentLocator.getByRole('textbox');
+    if (label) {
+      locator = parentLocator.getByLabel(label, { exact: true }).first();
+    }
+    super(page, locator, locator, parentLocator);
+  }
+}
+
+export class MacroEditor extends CodeEditor {
+  constructor(page: Page, parentLocator: Locator, label: string) {
+    const locator = parentLocator.getByLabel(label, { exact: true }).first();
+    super(page, locator, locator, parentLocator);
+  }
+
+  async expectValue(value: string) {
+    await expect(this.value).toHaveText(value.replace('\n', ''));
   }
 }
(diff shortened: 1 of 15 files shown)


packages/editor/src/components/widgets/code-editor/MacroArea.tsx
```diff
@@ -1,26 +1,34 @@
...
     <div className='script-area' {...focusWithinProps} tabIndex={1}>
       {isFocusWithin || browser.open ? (
         <>
-          <ResizableCodeEditor {...props} location={path} onMountFuncs={[setEditor, monacoAutoFocus]} macro={true} />
+          <ResizableCodeEditor
+            {...props}
+            location={path}
+            onMountFuncs={[setEditor, monacoAutoFocus]}
+            macro={true}
+            initHeight={areaRef.current?.offsetHeight}
```
ivy-lli, line 26: init monaco editor with the same height as the read mode output has

packages/editor/src/components/widgets/output/CardText.tsx
```diff
@@ -0,0 +1,59 @@
...
+    <output className='card-text' {...props}>
+      <CardLine parts={parts} />
+    </output>
+  );
+};
+
+export const CardArea = forwardRef<ElementRef<'output'>, CardTextProps>(({ value, ...props }, forwardRef) => {
+  const lines = splitNewLine(value ?? '').map(line => splitMacroExpressions(line));
+  return (
+    <output className='card-text' {...props} ref={forwardRef}>
```
weissreto, line 51: output is a standard thing?

packages/editor/src/utils/utils.test.ts
```diff
@@ -0,0 +1,29 @@
...
+  expect(splitMacroExpressions('abc')).toEqual(['abc']);
+  expect(splitMacroExpressions('abc <%= ivy.cms.co("hi") %>')).toEqual(['abc ', '<%= ivy.cms.co("hi") %>']);
+  expect(splitMacroExpressions('abc <%= ivy.cms.co("hi") %>, hi')).toEqual(['abc ', '<%= ivy.cms.co("hi") %>', ', hi']);
+  expect(splitMacroExpressions('abc <%= ivy.cms.co("hi") %>, hi <%=in.accept%>')).toEqual([
+    'abc ',
+    '<%= ivy.cms.co("hi") %>',
+    ', hi ',
+    '<%=in.accept%>'
+  ]);
+  expect(splitMacroExpressions('<%= ivy.cms.co("hi") %><%=in.accept%>')).toEqual(['<%= ivy.cms.co("hi") %>', '<%=in.accept%>']);
```
weissreto, line 28: devil test '<%= ivy.cms.co("hi %>")%>'

weissreto: Does not need to be fixed for the first version. ;-)

packages/editor/src/components/widgets/output/CardText.tsx
```diff
@@ -0,0 +1,59 @@
...
+};
+
+const CardOrText = ({ part }: { part: string }) => {
+  if (part.startsWith('<%=')) {
+    return <Card part={part} />;
+  }
+  return <span className='text'>{part}</span>;
+};
+
+const CardLine = ({ parts }: { parts: string[] }) => (
```
weissreto, line 31: Why not use an array direct here instead of wrapping it to an object?

weissreto (approved):
Good features. Some comments.

packages/editor/src/components/widgets/output/CardText.tsx
ivy-lli: Yes didn't know it either, but this is officially an HTML element.
The point why I need this one is that an input element cannot have sub-elements (at least as far as I know), but it needs to be an element to which we can link a label or the tests will no longer work. 
And by the way, it makes sense to me, that this is an output field...

integrations/standalone/tests/pageobjects/CodeEditor.ts
```diff
@@ -44,3 +32,28 @@ export class CodeEditor {
...
+  }
+}
+
+export class MacroArea extends ScriptInput {
+  async expectValue(value: string) {
+    await expect(this.value).toHaveText(value.replace('\n', ''));
+  }
+}
+
+export class MacroInput extends MacroArea {}
```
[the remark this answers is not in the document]
ivy-lli: I will change it a bit :)

packages/editor/src/utils/utils.test.ts
ivy-lli: I'm not sure if it is really a use case that we ever support this case in this UI (even if it's technical totally fine), but I think it is not worth to fix this until someone reports this as a bug.
Unless you have the regex for it right away? 😄 
But you're right a test for this does definitely not harm 👍

packages/editor/src/components/widgets/output/CardText.tsx
ivy-lli: This is simply not supported by react. React always needs a single object as parameter (called props).

Let's expect that we don't have an object: `const CardLine = (parts: string[]) => (`
Then how would the call of this look like in JSX? `<CardLine ['abc'] />`

In HTML and also JSX every attribute needs to be named and react does this with the object `props`, but then you can call the component with the props in every order.

packages/editor/src/utils/utils.test.ts
weissreto: You cannot solve such a thing with a simple Regex because it needs a state (inside string or not). Have a look at ch.ivyteam.ivy.scripting.expander.impl.TextTagExpander. But as said it is not a prio one.

packages/editor/src/components/widgets/output/CardText.tsx
weissreto: Ok, thank for the education.

PR #246, opened 2023-08-03
Fix reset functionality

ivy-lli:
- Fix broken reset functionality and add test for it
- Change MacroArea, MacroInput and ScriptInput fields -> save data not after every change only after focus loose
  - Reduce pressure on server (especially in the test case)
  - No nervous validation messages while write something (the validations are shown anyways in the monaco editor)

### DIFF
--- a/integrations/standalone/tests/pageobjects/Accordion.ts
+++ b/integrations/standalone/tests/pageobjects/Accordion.ts
@@ -27,6 +27,10 @@ export class Accordion extends Part {
     return new Tab(this.page, this.locator, label);
   }
 
+  resetButton() {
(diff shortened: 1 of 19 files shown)


integrations/standalone/tests/pageobjects/Accordion.ts
```diff
@@ -27,6 +27,10 @@ export class Accordion extends Part {
     return new Tab(this.page, this.locator, label);
   }
 
+  reset() {
```
weissreto, line 30: resetButton

weissreto (approved):
Bfff. Looks good but I'm not the expert here.

PR #247, opened 2023-08-03
XIVY-11703 Add trigger part to start request

ivy-lli:


weissreto (approved):
didn't understand everything. But those parts I understand look good.

PR #248, opened 2023-08-03
Screenshots

ivy-lli:
- Enable some screenshot tests
- Use pageobjects for screenshot tests
- Remove some utils
- Screenshot only accordion instead of full page

### DIFF
--- a/integrations/standalone/tests/pageobjects/Composite.ts
+++ b/integrations/standalone/tests/pageobjects/Composite.ts
@@ -60,4 +60,8 @@ export abstract class Composite {
   tags(): any {
     return new Tags(this.page, this.locator);
   }
+
+  currentLocator() {
+    return this.locator;
+  }
 }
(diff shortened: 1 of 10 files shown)


integrations/standalone/tests/pageobjects/Composite.ts
```diff
@@ -60,4 +60,8 @@ export abstract class Composite {
   tags(): any {
     return new Tags(this.page, this.locator);
   }
+
+  currentLocator() {
```
weissreto, line 64: locator() does not work because locator is already a property?

weissreto: then it is maybe better to rename the property and make it private. And use locator() outside of the class

weissreto (approved):
Looks good. 2 small comments

integrations/standalone/tests/pageobjects/Composite.ts
ivy-lli: Yes, and yes. I will change that but I think in a further PR as I have many open PRs which may result in merge conflicts 🙈

PR #250, opened 2023-08-03
Add signal end error pageobject tests

ivy-lli:


integrations/standalone/tests/integration/parts/signal-catch.ts
```diff
@@ -1,38 +1,46 @@
...
 
-export class SignalCatchTester implements PartTest {
-  constructor(private readonly makroSupport: boolean = false) {}
+class SignalCatch extends PartObject {
+  signal: Combobox;
+  attach: Checkbox;
 
-  partName() {
-    return 'Signal';
+  constructor(part: Part, private readonly makroSupport: boolean = false) {
```
weissreto, line 10: markoSupport -> attachToBusinessCaseSupport?

weissreto (approved):
One small comment

integrations/standalone/tests/integration/parts/signal-catch.ts
ivy-lli: This technically does not only add the checkbox, but also adds a macroinput to the singal, but as this is currently a combobox, I'm not sure how to implement that in the UI...

PR #255, opened 2023-08-10
XIVY-11703 Add start integration tests

ivy-lli:
Fix persist on start flag:
- Wrong config key
- Disabled if anonymous start is allowed

Add integration tests for request start:
- Request
- Trigger
- Task
- Case

### DIFF
--- a/integrations/standalone/tests/integration/parts/task.ts
+++ b/integrations/standalone/tests/integration/parts/task.ts
@@ -58,6 +58,7 @@ class Task extends PartObject {
   optionsSection: Section;
   skipTasklist: Checkbox;
   delay: ScriptInput;
+  persist: Checkbox;
   expirySection: Section;
   timeout: ScriptInput;
   error: Select;
@@ -68,7 +69,12 @@ class Task extends PartObject {
   codeSection: Section;
   code: ScriptArea;
 
-  constructor(part: Part, private readonly nameValue: string, private readonly errorValue: RegExp) {
+  constructor(
+    part: Part,
+    private readonly nameValue = 'test name',
+    private readonly errorValue = /f8/,
+    private readonly showPersist = false
+  ) {
     super(part);
     this.name = part.macroInput('Name');
     this.description = part.macroArea('Description');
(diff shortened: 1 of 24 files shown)


packages/protocol/src/data/workflow-data.ts
```diff
@@ -1,26 +1,26 @@
 import { WfActivatorType, WfFieldType, WfLevel } from './inscription';
 
-export const RESPONSIBLE_TYPE: Record<WfActivatorType, string> = {
+export const RESPONSIBLE_TYPE = {
   ROLE: 'Role',
   ROLE_FROM_ATTRIBUTE: 'Role from Attr.',
   USER_FROM_ATTRIBUTE: 'User from Attr.',
   DELETE_TASK: 'Nobody & delete'
-} as const;
+} as const satisfies Record<WfActivatorType, string>;
```
ivy-lli, line 8: This is something new in TS since 5.1 or 5

The problem with the Recode before was that the values were no longer known (ts forgots about it and only us string as value). But it gives you the typesafty, that no other key as on of WfActivatorType can be added to this object.

With satisfies this is also guaranteed, but without the loose of the specific values.

packages/protocol/src/data/part-data.ts
```diff
@@ -39,7 +39,7 @@ export type ConditionData = ElementAlternative;
 export interface TaskData {
   task: WfTask;
   tasks: WfTask[];
-  persist: boolean;
+  persistOnStart: boolean;
```
ivy-lli, line 42: Somehow this was renamed, so no longer fitted the config...
I work on something to make this stuff more "typo" or schema safe.

integrations/standalone/tests/integration/parts/task.ts
```diff
@@ -179,11 +205,12 @@ class Task extends PartObject {
...
-    super('Task', (part: Part) => new Task(part, options?.name ?? 'test name', options?.error ?? /f8/));
+  constructor(options?: { name?: string; error?: RegExp; persist?: boolean }) {
+    super('Task', (part: Part) => new Task(part, options?.name, options?.error, options?.persist));
   }
 }
 
 export const TaskTest = new TaskTester();
-
+export const StartRequestTaskTest = new TaskTester({ error: /EventAndGateway/, persist: true });
+export const TaskIntermediateTaskTest = new TaskTester({ error: /EventAndGateway/ });
```
weissreto, line 215: Seems not to be used in any tests yet?

weissreto (approved):
👀✔. I almost understand every change. It is getting better ... ;-)

integrations/standalone/tests/integration/parts/task.ts
ivy-lli: Ah yes. Forgot that. 
I thought it is easier if the test knows which parameters are needed for a specific element...

PR #256, opened 2023-08-10
Use more types form the inscription schema 🧙🏼‍♂️

ivy-lli:
This change has the effect, that more types are generated out of the inscription schema from the core.

### DIFF
--- a/packages/protocol/src/data/part-data.ts
+++ b/packages/protocol/src/data/part-data.ts
@@ -1,78 +1,34 @@
-import {
-  Data,
-  ElementAlternative,
-  ElementEMail,
-  ElementScript,
-  ScriptMapCode,
-  ScriptParameterizedMapCode,
-  StartPermission,
-  StartRequest,
-  WfCase,
-  WfTask
-} from './inscription';
+import { Data } from './inscription';
+import { ConfigData } from './inscription-data';
 
 export type NameData = Omit<Data, 'config'>;
 
-export type ProcessCallData = {
-  processCall: string;
-};
+export type ProcessCallData = Pick<ConfigData, 'processCall'>;
 
-export type DialogCallData = {
-  dialog: string;
-};
+export type DialogCallData = Pick<ConfigData, 'dialog'>;
 
-export type MappingData = ScriptMapCode;
-export type Mapping = MappingData['map'];
+export type CallData = Pick<ConfigData, 'call'>;
 
-export type CallData = {
-  call: MappingData;
-};
+export type EndPageData = Pick<ConfigData, 'page'>;
 
-export interface EndPageData {
-  page: string;
-}
+export type OutputData = Pick<ConfigData, 'output' | 'sudo'>;
 
-export type OutputData = ElementScript;
+export type ConditionData = Pick<ConfigData, 'conditions'>;
 
-export type ConditionData = ElementAlternative;
+export type TaskData = Pick<ConfigData, 'persistOnStart' | 'task' | 'tasks'>;
 
-export interface TaskData {
-  task: WfTask;
-  tasks: WfTask[];
-  persistOnStart: boolean;
-}
+export type CaseData = Pick<ConfigData, 'case'>;
 
-export interface CaseData {
-  case: WfCase;
-}
+export type StartData = Pick<ConfigData, 'input' | 'signature'>;
 
-export interface StartData {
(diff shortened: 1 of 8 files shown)


packages/protocol/src/data/part-data.ts
```diff
@@ -1,78 +1,34 @@
...
+export type ConditionData = Pick<ConfigData, 'conditions'>;
 
-export type ConditionData = ElementAlternative;
+export type TaskData = Pick<ConfigData, 'persistOnStart' | 'task' | 'tasks'>;
 
-export interface TaskData {
-  task: WfTask;
-  tasks: WfTask[];
-  persistOnStart: boolean;
-}
```
ivy-lli, line 43: Do no longer define our own types, but extract the needed configs from the ConfigData.
This has the benefit, that e.g. the ts typechecker fails if the `persist` flag was renamed to `persistOnStart`

packages/protocol/src/defaults/common-defaults.ts
```diff
@@ -131,6 +141,25 @@ export const DEFAULT_DATA: ElementData = {
...
+    throws: {} as ErrorDefinition,
+    link: '',
+    clientId: '',
+    operation: {} as SoapOperation,
+    properties: {},
+    exception: {} as SoapWsProcessException,
+    eventId: '',
+    method: 'GET',
+    response: {} as RestResponse,
+    target: {} as RestTarget
```
ivy-lli, line 163: this is the only downside of this change at the moment...
We need to tell TS about this default data which is currently not used. But with each finished element view we would need to do this anyway, so this is currently only a placeholder until it is replaced by the real defaults when we implement those parts.

packages/protocol/src/utils/type-helper.ts
```diff
@@ -9,3 +9,9 @@ export type ValuesAsUnionDeep<T> = {
 declare const brand: unique symbol;
 
 export type Brand<T, TBrand> = T & { [brand]: TBrand };
+
+export type UnionToIntersection<U> = (U extends any ? (k: U) => void : never) extends (k: infer I) => void ? I : never;
```
ivy-lli, line 13: This converts a type of unions to an intersected type, e.g.
`{name: string} | {description: string}` becomes to 
`{name: string} & {description: string}` or better 
`{name: string; description: string}`

packages/protocol/src/utils/type-helper.ts
```diff
@@ -9,3 +9,9 @@ export type ValuesAsUnionDeep<T> = {
...
+
+export type Prettify<T> = {
+  [K in keyof T]: T[K];
+} & {};
```
ivy-lli, line 17: typehelper to see the real type, e.g. if you hover the ConfigData from above you only see something like: `type ConfigData = ElementTaskEndPage & ElementTaskSwitchGateway & ElementSubProcessCall & ElementDatabase & ... 27 more ... & ElementSignalStartEvent`
with `Prettify<ConfigData>` you see something like `type ConfigData = {
    page: string;
    output: ScriptMapCode;
    case: WfCase;
    tasks: WfTask[];
    call: ScriptMapCode;
    processCall: string;
    cache: Cache;
    query: DbQuery;
    ... 34 more ...;
    attachToBusinessCase: boolean;
}`

not used currently, but really helpful if you try to write such ts types 🧙

weissreto: Wenn ich das richtig verstehe funktioniert das so:

Element1Data {
a:string,
b:string}
Element2Data {
c:string,
d:string}

ConfigData = {
a:string
b:string
c:string
d:string}

Element1InscriptionConfig {
a:string}
Element2InscriptionConfig {
c:string}

Man nimmt also alle Typen vom Protocol mixt die zusammen damit man einen Typ mit allen Properties hat und man nacher wieder einzelne Typen für die Inscription Masken erstellen kann die jedoch einzelne Properties aus dem Ursprünglichen Typen nicht mehr drin haben?

Warum macht man nicht für jeden Type eine Mapping Funktion die Protocol Objekt auf Inscription Object mapped anstelle hier mit Type Magic zu arbeiten, die nur wirkliche TS Experten verstehen?

Vielleicht hab ich auch einfach etwas nicht richtig verstanden.

weissreto (commented):
Siehe mein Kommentar

PR #259, opened 2023-08-14
Task part refactoring

ivy-lli:
- Pack Name, Description and Category into own component
  - Use it for Task, Case and Request part
- Refactor task part a bit to reduce ifs
  - Own components for StartRequest, Wait and WsStart Task part (and add those last two)


ivy-rew (approved):
nice generalization :+1:

PR #261, opened 2023-08-16
Remove CodeEditorUtil

ivy-lli:
The browser test is not really a beauty, but as this functionality is only tested here at the moment, I didn't invest into move this code into an own pageobject.


integrations/standalone/tests/pageobjects/Table.ts
```diff
@@ -115,8 +115,10 @@ export class Cell {
   async edit(value: string) {
     const expression = this.locator.getByRole('textbox');
     await expect(expression).toHaveAttribute('aria-expanded', 'false');
-    await expression.click();
-    await CodeEditorUtil.type(this.page, value);
+    const code = new CodeEditor(this.page, expression, expression, this.locator);
+    await code.focus();
+    await code.clearContent();
+    await this.page.keyboard.type(value);
```
weissreto, line 121: code.fill(value) does something like this already

integrations/standalone/tests/mock/browser.spec.ts
```diff
@@ -23,36 +23,45 @@ test.describe('Script browser', () => {
...
 
+    const category = task.macroInput('Category');
     await assertCodeHidden(page);
-    await page.getByLabel('Category').click();
+    await category.focus();
     await assertCodeVisible(page);
 
-    await CodeEditorUtil.fill(page, 'test 123 zag');
-    await page.getByRole('code').dblclick();
+    await page.keyboard.type('test 123 zag');
```
weissreto, line 31: category.fill

weissreto (changes requested):
I would try to hide thinks like page.keyboard.type also into the CodeEditor

integrations/standalone/tests/pageobjects/Table.ts
ivy-lli: fill does make a blur at the end

integrations/standalone/tests/mock/browser.spec.ts
ivy-lli: fill does make a focus, clear, type and blur. I only want to write...

integrations/standalone/tests/pageobjects/Table.ts
ivy-lli: I will make an own impl of the code editor for ScriptCell.

weissreto (approved):
🙏

PR #267, opened 2023-08-22
Fix windows environment

ivy-lli:
- Downgrade fantasticon to v1.2.3: v2.0.0 uses newer version of glob, which causes path problems under windows
- Add mock start for windows


ivy-lli:
Hi @ivy-edp 
Can you quickly test if these fixes work?

ivy-edp:
The fixes work

PR #269, opened 2023-08-23
XIVY-11710 Add error throw part to error end

ivy-lli:


weissreto (commented):
Seems to be good. As far as I understand the code

PR #271, opened 2023-08-25
XIVY-11710 Synch name of error end with error

ivy-lli:
Feedback of bb: Error End display name should update to the current error code (if was the same)


weissreto (approved):
✔

PR #272, opened 2023-08-25
XIVY-11711 Add query part to db

ivy-lli:
Full UI support of DB query view.
Missing (will follow in a next PR):
- unit tests
- integration tests
- validations

Screen Recording 2023-08-25 at 14 39 34

### DIFF
--- a/packages/editor/src/components/parts/mail/MailHeaderPart.test.tsx
+++ b/packages/editor/src/components/parts/mail/MailHeaderPart.test.tsx
@@ -82,6 +82,6 @@ describe('MailHeaderPart', () => {
     expect(data.config.headers.bcc).toEqual('');
     expect(data.config.headers.to).toEqual('');
     expect(data.config.failIfMissingAttachments).toBeFalsy();
-    expect(data.config.exceptionHandler).toEqual('ivy:error:email');
+    expect(data.config.exceptionHandler).toEqual('');
   });
 });
--- a/packages/editor/src/components/parts/query/database/DbExceptionHandler.tsx
+++ b/packages/editor/src/components/parts/query/database/DbExceptionHandler.tsx
@@ -1,0 +1,17 @@
+import { Collapsible } from '../../../widgets';
+import { ExceptionSelect, IGNROE_EXCEPTION } from '../../common';
+import { useQueryData } from '../useQueryData';
+import { IVY_EXCEPTIONS } from '@axonivy/inscription-protocol';
+
+export const DbExceptionHandler = () => {
+  const { config, defaultConfig, updateException } = useQueryData();
+  return (
+    <Collapsible label='Error' defaultOpen={config.exceptionHandler !== defaultConfig.exceptionHandler}>
+      <ExceptionSelect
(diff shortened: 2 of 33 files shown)


packages/protocol/src/data/inscription.ts
```diff
@@ -25,14 +25,19 @@ export interface Inscription {
...
   inscriptionSaveRequest: InscriptionSaveRequest;
   callableStart: CallableStart[];
   scriptingDataArgs: ScriptingDataArgs;
   variableInfo: VariableInfo;
   errorMeta: ErrorMeta[];
   eventCodeMeta: EventCodeMeta[];
   roleMeta: RoleMeta[];
-  connectorRef: ConnectorRef;
+  connectorRef: ConnectorRef[];
+  string: string[];
```
weissreto, line 37: string is very common. Better to use tables?

weissreto: and databases?

weissreto (approved):
Cool. Some minor thoughts

packages/editor/src/components/parts/mail/MailHeaderPart.test.tsx
```diff
@@ -82,6 +82,6 @@ describe('MailHeaderPart', () => {
     expect(data.config.headers.bcc).toEqual('');
     expect(data.config.headers.to).toEqual('');
     expect(data.config.failIfMissingAttachments).toBeFalsy();
-    expect(data.config.exceptionHandler).toEqual('ivy:error:email');
+    expect(data.config.exceptionHandler).toEqual('');
```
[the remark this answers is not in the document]
ivy-lli: Because both db query and email do use the exceptionHandler parameter, but both with another default value (ivy:error:email for mail, and ivy:error:database for db). 
Currently the test default value was on ivy:error:email, but I now set it to an empty string.
As this has only some testing effects I thought I makes no difference, but if you like I can add the correct default via parameter

packages/editor/src/components/parts/query/database/DbExceptionHandler.tsx
```diff
@@ -0,0 +1,25 @@
+import { Collapsible, Select, SelectItem } from '../../../widgets';
+import { IGNROE_EXCEPTION } from '../../common';
+import { useQueryData } from '../useQueryData';
+import { IVY_EXCEPTIONS } from '@axonivy/inscription-protocol';
+import { useMemo } from 'react';
+
+export const DbExceptionHandler = () => {
```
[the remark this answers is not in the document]
ivy-lli: Ahh I thought the db only supports the default code or ignoring. But yes it seems it also supports pid references. So I will reuse the exiting component 👍

packages/protocol/src/data/inscription.ts
ivy-lli: This is the generated code from the inscription schema.
I think it calls the parameter like the type, and as I return a list of strings (e.g. as db names) it is called string here...
But this Inscription type here is anyways only here, so the types are correctly exported (or generated) I think. So it is not directly used.

PR #286, opened 2023-09-14
Prettier ignore inscription.ts file

ivy-lli:


packages/protocol/src/data/inscription.ts
```diff
@@ -1,11 +1,12 @@
 /* eslint-disable */
+// prettier-ignore
```
ivy-lli, line 2: This line will disable prettier for this file

ivy-lli:
@ivy-rew Somehow the ordering of the schema seems not to be the same after every change...

ivy-rew (commented):
:+1:

ivy-rew:
> @ivy-rew Somehow the ordering of the schema seems not to be the same after every change...

yep, I don't like that either. In theory I can enforce an order in the raw schema file ... let's make a task to address if it really hurts.

PR #290, opened 2023-09-18
XIVY-11707 Add Process Inscription View

ivy-edp:
- Added Business & WS Process Inscription View (with IT)
- Added Name view
- Added Process Data view (with Unittests)
- Added Permissions view (with Unittests)
- Added Web Service Process view (with Unittests)

(ToDo: Validation)


packages/editor/src/components/editors/NameEditor.tsx
```diff
@@ -3,8 +3,8 @@ import { memo } from 'react';
 import { useNamePart } from '../parts';
 import InscriptionEditor from './InscriptionEditor';
 
-const NameEditor = memo((props: { icon: IvyIcons; hideTags?: boolean }) => {
-  const name = useNamePart({ hideTags: props.hideTags });
+const NameEditor = memo((props: { icon: IvyIcons; hideTags?: boolean; disableName?: boolean }) => {
+  const name = useNamePart({ hideTags: props.hideTags, disableName: props.disableName });
```
ivy-lli, line 7: I think this change here is not needed, as noone uses this NameEditor with the disabledName right? This is only the case for the process editor and those don't use the NameEditor.

packages/editor/src/components/parts/name/NamePart.test.tsx
```diff
@@ -31,6 +31,12 @@ describe('NamePart', () => {
     expect(screen.queryByText('Tags')).not.toBeInTheDocument();
   });
 
+  test('disable name', async () => {
```
ivy-lli, line 34: 👍

packages/editor/src/components/parts/name/NamePart.tsx
```diff
@@ -12,20 +12,26 @@ export function useNamePart(options?: { hideTags?: boolean }): PartProps {
...
 
   const nameField = useFieldset();
   const descriptionField = useFieldset();
   return (
     <>
       <Fieldset label='Display name' {...nameField.labelProps}>
-        <Textarea maxRows={3} value={data.name} onChange={change => update('name', change)} {...nameField.inputProps} />
+        <Textarea
+          maxRows={3}
+          disabled={props.disableName ? true : false}
```
ivy-lli, line 30: you can also do `!!props.disableName` 
this is a duplicated negation, so if it is true or false, it will be again the same
if it is undefined -> first negation: true -> second negation: false

packages/editor/src/components/parts/permissions/PermissionsPart.test.tsx
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  test('full data', async () => {
+    renderPart({
+      permissions: {
+        view: { allowed: false }
+      }
+    });
+    expect(screen.getByLabelText('Allow all workflow users to view the process on the Engine')).not.toBeChecked();
+  });
```
ivy-lli, line 37: move this test besides the empty data test 👍

packages/editor/src/components/parts/permissions/PermissionsPart.test.tsx
```diff
@@ -0,0 +1,68 @@
...
+    assertState('configured', {
+      permissions: {
+        view: { allowed: false }
+      }
+    });
+
+    //not really necessary?
+    assertState('error', undefined, { path: 'permissions.cause', message: '', severity: 'ERROR' });
+    assertState('warning', undefined, { path: 'permissions.error', message: '', severity: 'WARNING' });
+  });
```
ivy-lli, line 50: yes it is not, so remove it :)

packages/protocol/src/defaults/common-defaults.ts
```diff
@@ -186,6 +187,11 @@ export const DEFAULT_WSREQUEST_DATA: WsRequestData = {
   properties: {}
 } as const;
 
+export const DEFAULT_PROCESS_DATA: ProcessConfig = {
+  data: '',
+  permissions: { view: { allowed: true } }
+};
```
ivy-lli, line 193: I think this is not necessary, as the DEFAULT_WSPROCESS_DATA already defines this default values

packages/editor/src/components/parts/permissions/PermissionsPart.tsx
```diff
@@ -0,0 +1,28 @@
...
+import { PathContext, useValidations } from '../../../context';
+import { Checkbox } from '../../../components/widgets';
+import { usePermissionsData } from './usePermissionsData';
+import { PermissionsData } from '@axonivy/inscription-protocol';
+
+export function usePermissionsPart(): PartProps {
+  const { config, defaultConfig, initConfig, reset } = usePermissionsData();
+  const compareData = (data: PermissionsData) => [data];
+  const validation = useValidations(['permissions']);
+  const state = usePartState(compareData(defaultConfig), compareData(config), validation);
```
ivy-lli, line 11: this part has no validations, so you can simply insert an empty array into usePartState

packages/editor/src/components/parts/permissions/PermissionsPart.tsx
```diff
@@ -0,0 +1,28 @@
...
+  const dirty = usePartDirty(compareData(initConfig), compareData(config));
+  return { name: 'Permissions', state, reset: { dirty, action: () => reset() }, content: <PermissionsPart /> };
+}
+
+const PermissionsPart = () => {
+  const { config, update } = usePermissionsData();
+
+  return (
+    <PathContext path='permissions'>
```
ivy-lli, line 20: You don't need a path context here, as you have no validation messages nor a script context

packages/editor/src/components/parts/process-data/ProcessDataPart.test.tsx
```diff
@@ -0,0 +1,61 @@
...
+    });
+    expect(result.current.state.state).toEqual(expectedState);
+  }
+
+  test('full data', async () => {
+    renderPart({
+      data: 'screenshot.project.Order'
+    });
+    expect(screen.getByLabelText('Data Class')).toHaveValue('screenshot.project.Order');
+  });
```
ivy-lli, line 35: move this test beside the empyt test

packages/editor/src/components/parts/process-data/ProcessDataPart.test.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+  test('configured', async () => {
+    assertState('empty');
+    assertState('configured', {
+      data: 'screenshot.project.Order'
+    });
+
+    assertState('error', undefined, { path: 'data.cause', message: '', severity: 'ERROR' });
+    assertState('warning', undefined, { path: 'data.error', message: '', severity: 'WARNING' });
```
ivy-lli, line 44: one test is enough (warning or error)

packages/editor/src/components/parts/process-data/ProcessDataPart.tsx
```diff
@@ -0,0 +1,45 @@
...
+}
+
+const ProcessDataPart = () => {
+  const { config, update } = useProcessDataData();
+
+  const dataClassField = useFieldset();
+
+  return (
+    <PathContext path='data'>
+      <PathFieldset label='Data Class' {...dataClassField.labelProps} path='data'>
```
ivy-lli, line 29: remove this path context. or you will endup with a path context "data.data" which is wrong. The PathFieldset is just fine.
If you have the error that you can not put to components in the same return level you can simply use a jsx fragmet around it 
```js
<>
  <PathFieldset ... />
  <MessageText ... />
</>
```

packages/editor/src/components/parts/web-service-process/WebServiceProcessPart.tsx
```diff
@@ -0,0 +1,47 @@
...
+      </PathFieldset>
+      <PathFieldset label='Authentication' {...authField.labelProps} path='wsAuth'>
+        <Radio
+          value={config.wsAuth}
+          onChange={change => update('wsAuth', change)}
+          items={[
+            { label: 'None/Container', value: 'NONE' },
+            { label: 'WS Security', value: 'WS_SECURITY' },
+            { label: 'HTTP Basic', value: 'HTTP_BASIC' }
+          ]}
```
ivy-lli, line 41: this values here are known via the protocol. So I would suggest that we should at least add some type safety to them should they ever change.

Put into the workflow-data.ts file in the protocol/src/data an entry: 
```ts
export const WS_AUTH_TYPE = {
  NONE: 'None/Container',
  WS_SECURITY: 'WS Security',
  HTTP_BASIC: 'HTTP Basic'
} as const satisfies Record<WsAuth, string>;
```

then you can reuse them here by e.g.:
```ts
Object.entries(WS_AUTH_TYPE).map(([value, label]) => ({ label, value }))
```

ivy-lli (changes requested):
Nice! Some small improvement requests :)

packages/editor/src/components/parts/web-service-process/WebServiceProcessPart.tsx
ivy-edp: i had to implement it like this because otherwise "change" was not happy with the type. i always had the following message: Argument of type 'string' is not assignable to parameter of type 'WsAuth'.

onChange={(change) => update('wsAuth', change)}
items={Object.entries(WS_AUTH_TYPE).map(([value, label]) => ({
   label,
   value: value as keyof typeof WS_AUTH_TYPE
}))}

ivy-lli: or simply do: `onChange={change => update('wsAuth', change as string)}`

ivy-lli (approved):
Perfect!

PR #294, opened 2023-09-20
XIVY-11707 Add HTM/Sub Process Inscription+ Bugfix

ivy-edp:


integrations/standalone/tests/integration/parts/html-dialog-permissions.ts
ivy-lli, line 1: Isn't this the same as the normal process permission?

ivy-edp: Unlike the normal permission, I had to change the assertFill to expectChecked instead of expectUnchecked and the assertClear to expectUnchecked instead of expectChecked, since the permission is not selected by default in the HTML dialog. Should I maybe reuse the already existing permissions.ts and work with a flag called "defaultIsChecked"?

ivy-lli: Ahh I wasn't aware of that. 
Yes I would prefer that, because it is technical the same view

integrations/standalone/tests/integration/other/callable-sub-process.spec.ts
```diff
@@ -0,0 +1,36 @@
+import { test } from '@playwright/test';
+import { InscriptionView } from '../../pageobjects/InscriptionView';
+import { NameTestWithDisabledName, runTest } from '../parts';
+import { CreateProcessResult, createProcess } from '../../glsp-protocol';
+import { PermissionsTest } from '../parts/permissions';
+import { ProcessDataTest } from '../parts/process-data';
```
ivy-lli, line 6: Add these two tests to the parts index.ts file, then you can import them via '../parts'

integrations/standalone/tests/integration/parts/permissions.ts
```diff
@@ -4,27 +4,38 @@ import { Checkbox } from '../../pageobjects/Checkbox';
...
+    if (this.defaultIsChecked) {
+      await this.viewable.expectChecked();
+    } else {
+      await this.viewable.expectUnchecked();
+    }
   }
 }
 
-export const PermissionsTest = new NewPartTest('Permissions', (part: Part) => new Permissions(part));
+export const PermissionsTest = (defaultIsChecked: boolean = true) =>
```
ivy-lli, line 40: if the default value is good enough you don't need to define the type: `(defaultIsChecked = true)`

ivy-lli (approved):
👍

ivy-lli (approved):

PR #298, opened 2023-09-27
Remove-copy-pasts

ivy-lli:
- Remove strange copy paste tests
- XIVY-11704 Add link to ws process

image


ivy-edp (commented):
I don't know now what the reason for these duplicates was in the first place but looks good 👍

PR #300, opened 2023-09-28
XIVY-11713 Add rest request target part

ivy-lli:
UI for Rest Target definitions:
image

Supported:
- Show all known clients with actions:
   - Jump to rest config
   - Create new rest config
- Select resource of known openapi resources
   - Or if disabled or if now openapi -> select method and enter a path
- Add parameters (query or path) and show all known parameters of openapi
   - OpenApi parameters are disabled (only expression is writable)
- Add headers (with all known headers with openapi headers)
   - Accept has it's own field, as it is not deletable, and it is the only header which has some known possible values
- Add properties (with known properties)

Missing:
- Target URL calculation
- Body part
- Unit tests
- Integration tests
- Validations


packages/editor/src/components/parts/rest-request/rest-target/parameters.ts
```diff
@@ -0,0 +1,56 @@
...
+
+export namespace Parameter {
+  export function of(parameters: RestParameter[], configs: ScriptMappings, kind: ParameterKind): Parameter[] {
+    const params = parameters.map<Parameter>(param => {
+      return {
+        kind,
+        name: param.name,
+        expression: '',
+        known: true,
+        doc: `${param.doc}${param.required ? '\n* requried' : ''}`,
```
ivy-rew, line 22: required

ivy-rew: show me first?

ivy-rew (approved):

PR #303, opened 2023-10-05
XIVY-11705 Add StartPart for ProgramInterface

ivy-edp:
Questions:
- Is the naming and generally the made folder structure for the start-part ok? (Top folder "program-start", subfolders "Activity" and "Event")
- Are the Unittests for the StartPart of the ProgramInterface (Activity) ok? At the moment i am not really testing the program and timeout section (currently I have not found a good way to access these with e.g. en label operators). Should I separate the program and timeout part of the StartPart component to make it easier to test?
- At the moment there is no test for the JavaClassSelector, because it is very small and the basic functionality is already tested in the StartPart.test.tsx. Is that OK?
- Is the validation and compareData in both StartParts correctly implemented?
- I had to adjust the MailHeaderTest because I have changed the default exceptionHandler-Value. Is this correct or is there a better way?


packages/editor/src/components/parts/mail/MailHeaderPart.test.tsx
```diff
@@ -82,6 +82,6 @@ describe('MailHeaderPart', () => {
     expect(data.config.headers.bcc).toEqual('');
     expect(data.config.headers.to).toEqual('');
     expect(data.config.failIfMissingAttachments).toBeFalsy();
-    expect(data.config.exceptionHandler).toEqual('');
+    expect(data.config.exceptionHandler).toEqual('ivy:error:program:exception');
```
ivy-lli, line 85: See answers

packages/editor/src/components/parts/program-start/JavaClassSelector.tsx
```diff
@@ -0,0 +1,30 @@
...
+  type: Type;
+};
+
+const JavaClassSelector = ({ config, onChange, type }: JavaClassSelectorProps) => {
+  const { context } = useEditorContext();
+  const javaClassItems = [
+    ...useMeta('meta/program/types', { type: type, context }, []).data.map<ComboboxItem>(javaClass => {
+      return { value: javaClass.fullQualifiedName };
+    })
+  ];
```
ivy-lli, line 19: you already get here an array back from the meta request and you don't have to destructure it and "repack" it into an array again. Simply do:
`const javaClassItems = useMeta('meta/program/types', { type: type, context }, []).data.map<ComboboxItem>(javaClass => ({ value: javaClass.fullQualifiedName }));`

The destructure / "repack" is a cool syntax if you directly want to add multiple stuff into an array.

packages/editor/src/components/parts/program-start/activity/StartPart.tsx
```diff
@@ -0,0 +1,66 @@
...
+    reset: { dirty, action: () => reset() },
+    content: <StartPart />
+  };
+}
+
+const StartPart = () => {
+  const { config, update, updateTimeout } = useProgramInterfaceStartData();
+
+  const errorFieldset = useFieldset();
+  const timeoutSeconsFieldset = useFieldset();
```
ivy-lli, line 28: Secons -> Seconds

packages/editor/src/components/parts/program-start/activity/StartPart.tsx
```diff
@@ -0,0 +1,66 @@
...
+  const timeoutErrorFieldset = useFieldset();
+
+  return (
+    <>
+      <JavaClassSelector config={config.javaClass} onChange={change => update('javaClass', change)} type='ACTIVITY' />
+
+      <PathCollapsible label='Program' path='exceptionHandler' defaultOpen={true}>
```
ivy-lli, line 35: implement default open correctly -> only if the configs are not default stuff

packages/editor/src/components/parts/program-start/activity/StartPart.tsx
```diff
@@ -0,0 +1,66 @@
...
+          <ExceptionSelect
+            value={config.exceptionHandler}
+            onChange={change => update('exceptionHandler', change)}
+            staticExceptions={[IVY_EXCEPTIONS.programException, IVY_EXCEPTIONS.ignoreException]}
+            inputProps={errorFieldset.inputProps}
+          />
+        </PathFieldset>
+      </PathCollapsible>
+
+      <PathCollapsible label='Timeout' path='timeout' defaultOpen={config.timeout !== undefined && config.timeout.seconds.length > 0}>
```
ivy-lli, line 46: better compare the config with the default config. Then it will still work if these defaults change sometimes and you also cover the error not only the seconds.

packages/editor/src/components/parts/program-start/activity/StartPart.test.tsx
```diff
@@ -0,0 +1,69 @@
...
+      javaClass: 'Test',
+      exceptionHandler: '>> Ignore Exception',
+      timeout: {
+        seconds: '',
+        error: 'ivy:error:program:timeout',
+        action: 'NOTHING',
+        duration: ''
+      }
+    });
+    expect(screen.getByLabelText('Java Class')).toHaveValue('Test');
```
ivy-lli, line 35: See answers

packages/editor/src/components/parts/program-start/activity/StartPart.test.tsx
```diff
@@ -0,0 +1,69 @@
...
+      wrapperProps: { data: data && { config: data }, validations: validation && [validation] }
+    });
+    expect(result.current.state.state).toEqual(expectedState);
+  }
+
+  test('configured', async () => {
+    assertState('empty');
+    assertState('configured', {
+      javaClass: 'Bla'
+    });
```
ivy-lli, line 49: I would put this on the same line to save space `assertState('configured', { javaClass: 'Bla' });`

And test other values too. I mean you currently only check that if the javaClass changes the accordion state changes to configured. What about the timeout? It's very easy to quickly check that your compareData function in the startpart is correctly implemented by check at least one of the configs in "timeout" and "exceptionHandler".

packages/editor/src/components/parts/program-start/activity/StartPart.test.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+    assertState('error', undefined, { path: 'javaClass.cause', message: '', severity: 'ERROR' });
+    assertState('warning', undefined, { path: 'exceptionHandler.error', message: '', severity: 'WARNING' });
```
ivy-lli, line 52: Check that the messages on the timeout also be catched by you're validation logic

packages/editor/src/components/parts/program-start/activity/StartPart.test.tsx
```diff
@@ -0,0 +1,69 @@
...
+        javaClass: 'Test'
+      }
+    };
+    const view = renderHook(() => useProgramInterfaceStartPart(), {
+      wrapperProps: { data, setData: newData => (data = newData) }
+    });
+    expect(view.result.current.reset.dirty).toEqual(true);
+
+    view.result.current.reset.action();
+    expect(data.config?.javaClass).toEqual('');
```
ivy-lli, line 67: check that the timeout and exceptionHandler configs are also correclty reseted by your code

packages/editor/src/components/parts/program-start/event/StartPart.tsx
```diff
@@ -1,30 +1,30 @@
...
-  return { name: 'Start', state, reset: { dirty, action: () => reset() }, content: <StartPart /> };
+  return {
+    name: 'Start',
+    state,
+    reset: { dirty, action: () => reset() },
+    content: <StartPart />
+  };
 }
 
 const StartPart = () => {
```
ivy-lli, line 22: name this and the file also like the hook -> ProgramStartPart

packages/editor/src/components/parts/program-start/activity/StartPart.tsx
```diff
@@ -0,0 +1,66 @@
...
+  const dirty = usePartDirty(compareData(initConfig), compareData(config));
+  return {
+    name: 'Start',
+    state,
+    reset: { dirty, action: () => reset() },
+    content: <StartPart />
+  };
+}
+
+const StartPart = () => {
```
ivy-lli, line 24: name this part like the hook -> ProgramInterfaceStartPart

packages/editor/src/components/parts/program-start/event/StartPart.test.tsx
ivy-lli, line 1: improve this test according to the comments on the interface startpart test

packages/editor/src/components/parts/program-start/event/useStartData.ts
```diff
@@ -31,6 +31,8 @@ export function useStartData(): ConfigDataContext<ProgramStartData> & {
       produce(draft => {
         draft.javaClass = config.initConfig.javaClass;
         draft.permission = config.initConfig.permission;
+        draft.exceptionHandler = config.initConfig.exceptionHandler;
+        draft.timeout = config.initConfig.timeout;
```
ivy-lli, line 35: Why does this also reset the exceptionHandler and the timeout? Those fields are not supported by the ProgrammStartEvent right? At least those fields seems not to be on the UI...

packages/protocol/src/defaults/common-defaults.ts
```diff
@@ -198,6 +197,17 @@ export const DEFAULT_REST_RESPONSE_DATA: RestResponseData = {
   }
 } as const;
 
+export const DEFAULT_PROGRAM_INTERFACE_START_DATA: ProgramInterfaceStartData = {
+  javaClass: '',
+  exceptionHandler: 'ivy:error:program:exception',
```
ivy-lli, line 202: see answers

ivy-lli:
Answers:
- I think i would name it just "program" as the intermediate part is not called "Start" but should also be inside of this folder.
- See review comments. Yes I would test this stuff. Not the fuctionality of the components, but that they show the correct value. If it makes testing easier I think it is ok to split the UI into smaller components and test those. But maybe it is not needed. There are other ways that select an element by label e.g. by the index. And for some components like the select there are utils to test them.
- It's ok for me that you don't test the select stuff. What you can test is the meta stuff: Check that the given meta's are correctly displayed as options or what happens if a value is given which is not a known option. (have a look at e.g. the ExceptionSelect test how to do this).
- I have no idea. You know those elements and their config much better I think! But if you have a look at the review comments you can test those logics a lot better. 
- This is only because you defined the "exceptionHandler" value in the common-defaults. As this is a value which is used by many elements and every element uses a different default value for it, I would leave it as it is now -> blank. So you don't have to define this in your defaults (see how the mail does it) or define it blank.

ivy-lli (changes requested):

packages/editor/src/components/parts/program-start/activity/StartPart.tsx
ivy-edp: Okay, I can do that, but in the already existing component this collapsible is also always open no matter if the value in the error is default or not or if the tab in was closed before.  So if I think about it, no collapsible would be necessary at all?

ivy-lli: Hmm nice to know. But I still would vote to implement it to be closed if nothing special is configured. 
But ok for me if you leave it as it is for now :)

packages/editor/src/components/parts/event/EventPart.tsx
```diff
@@ -0,0 +1,80 @@
...
+      <PathFieldset label='Event ID' path='eventId' {...eventIdFieldset.labelProps}>
+        <ScriptInput
+          value={config.eventId}
+          onChange={change => update('eventId', change)}
+          type={IVY_SCRIPT_TYPES.NUMBER}
+          {...eventIdFieldset.inputProps}
+        />
+      </PathFieldset>
+
+      <PathCollapsible label='Expiry' path='timeout' defaultOpen={config.timeout !== undefined && config.timeout.duration.length > 0}>
```
ivy-lli, line 50: same problem as before -> compare all fields for defaultOpen

packages/editor/src/components/parts/event/EventPart.tsx
ivy-lli, line 1: Why didn't you place the EventPart right beside the other start parts?
This is why I supposed to rename the `program-start` folder to just `program`, that you can also place this EventPart in it...

Then you don't need to have the javaClassSelector in the common part, but also in this `program` dir

packages/editor/src/components/parts/program/event/ProgramStartPart.test.tsx
```diff
@@ -17,7 +17,10 @@ describe('StartPart', () => {
 
   test('empty data', async () => {
     renderPart();
+    await SelectUtil.assertEmpty();
+    //await SelectUtil.assertOptionsCount(1);
```
ivy-lli, line 21: comment?

packages/editor/src/components/parts/program/event/ProgramStartPart.test.tsx
```diff
@@ -41,19 +48,15 @@ describe('StartPart', () => {
 
   test('configured', async () => {
     assertState('empty');
-    assertState('configured', {
-      javaClass: 'Bla'
-    });
+    assertState('configured', { javaClass: 'Bla', permission: { anonymous: true } });
```
ivy-lli, line 51: Do a separate check for the permission!
Imagine that your code which checks this state is wrong and the permission is not compared this would never fail because of the javaclass

packages/editor/src/components/parts/program/event/ProgramStartPart.test.tsx
ivy-edp: Yes I have left this line as a comment as I don't know yet what is the correct value for assertOptionsCount as I am currently not getting any data from the backend

packages/editor/src/components/editors/event/intermediate.tsx
```diff
@@ -4,7 +4,7 @@ import { ElementType } from '@axonivy/inscription-protocol';
 import { memo, ReactNode } from 'react';
 import InscriptionEditor from '../InscriptionEditor';
 import { useCasePart, useEndPagePart, useNamePart, useOutputPart, useTaskPart } from '../../../components/parts';
-import { useEventPart } from '../../../components/parts/event/EventPart';
+import { useEventPart } from '../../parts/program/intermediate/EventPart';
```
ivy-lli, line 7: you can import 'useEventPart' from '../../../components/parts'

ivy-lli (approved):
Looks good 👍

PR #312, opened 2023-10-09
XIVY-11705 Add EditorPart

ivy-edp:
First simple version of the EditorPart.
I haven't really separated the component yet, nor have I created any unit or integration tests. I think since it is not 100% known how or what data I will get from the backend I should not invest too much time in testing and separating the current solution. what do you think?
I also tried to format the multiline-label-text using linebreaks similar to the current version but I'm not satisfied yet. The current texts are separated quite randomly at certain places with linebreaks, so the separation-expression for these texts could become extremely long and unattractive. maybe i could skip this part at the moment as well.

### DIFF
--- a/packages/editor/src/components/parts/program/editor/Editor.css
+++ b/packages/editor/src/components/parts/program/editor/Editor.css
@@ -1,0 +1,11 @@
+.info-text {
+  display: flex;
+  flex-direction: column;
+  padding-inline: var(--size-1);
+  gap: var(--size-1);
+  font-size: 12px;
+}
+
+.info-text p {
+  margin: 0;
+}
--- a/packages/editor/src/components/parts/program/editor/EditorPart.test.tsx
+++ b/packages/editor/src/components/parts/program/editor/EditorPart.test.tsx
@@ -1,0 +1,68 @@
+import { DeepPartial, render, renderHook, screen } from 'test-utils';
+import { EditorData, ElementData, InscriptionValidation } from '@axonivy/inscription-protocol';
+import { PartStateFlag } from '../../../editors';
+import { useEditorPart } from './EditorPart';
+
+const Part = () => {
+  const part = useEditorPart();
+  return <>{part.content}</>;
+};
+
+describe('EditorPart', () => {
+  function renderPart(data?: DeepPartial<EditorData>) {
+    render(<Part />, {
+      wrapperProps: {
+        data: data && { config: data },
+        meta: {
+          widgets: [
+            { text: 'Path of directory to scan', multiline: false },
+            { multiline: false },
+            { text: 'Multiline-Text', multiline: true }
+          ]
+        }
+      }
+    });
+  }
+
+  test('empty data', async () => {
+    render(<Part />);
+    expect(screen.getByTitle('There is no editor for this bean')).toBeInTheDocument();
+  });
+
+  test('full data', async () => {
+    renderPart({
+      userConfig: '123'
+    });
+    await screen.findByText('Path of directory to scan');
+    expect(screen.getByDisplayValue('123')).toBeInTheDocument();
+  });
+
+  function assertState(expectedState: PartStateFlag, data?: DeepPartial<EditorData>, validation?: InscriptionValidation) {
+    const { result } = renderHook(() => useEditorPart(), {
+      wrapperProps: { data: data && { config: data }, validations: validation && [validation] }
+    });
+    expect(result.current.state.state).toEqual(expectedState);
+  }
+
+  test('configured', async () => {
+    assertState('empty');
+    assertState('configured', {
+      userConfig: 'test123'
+    });
+  });
+
+  test('reset', () => {
+    let data: DeepPartial<ElementData> = {
+      config: {
+        userConfig: 'test'
+      }
+    };
+    const view = renderHook(() => useEditorPart(), {
+      wrapperProps: { data, setData: newData => (data = newData) }
+    });
+    expect(view.result.current.reset.dirty).toEqual(true);
+
+    view.result.current.reset.action();
+    expect(data.config?.userConfig).toEqual('');
+  });
+});
--- a/packages/editor/src/components/parts/program/editor/EditorPart.tsx
+++ b/packages/editor/src/components/parts/program/editor/EditorPart.tsx
@@ -1,0 +1,76 @@
+import { EditorData, Label, Script, Text, Widget } from '@axonivy/inscription-protocol';
+import { PartProps, usePartDirty, usePartState } from '../../../editors';
+import { useEditorContext, useMeta, useValidations } from '../../../../context';
+import { useEditorData } from './useEditorData';
+import { Input, MessageText, ScriptInput } from '../../../../components/widgets';
+import './Editor.css';
+
+export function useEditorPart(): PartProps {
+  const { config, defaultConfig, initConfig, reset } = useEditorData();
+  const compareData = (data: EditorData) => [data.userConfig];
+  const validation = useValidations(['userConfig']);
+  const state = usePartState(compareData(defaultConfig), compareData(config), validation);
+  const dirty = usePartDirty(compareData(initConfig), compareData(config));
+  return {
+    name: 'Editor',
+    state,
+    reset: { dirty, action: () => reset() },
+    content: <EditorPart />
+  };
+}
+
+const EditorPart = () => {
+  const { config, update } = useEditorData();
+  const { context } = useEditorContext();
+  const editorItems = useMeta('meta/program/editor', { context, type: config.javaClass }, []).data;
+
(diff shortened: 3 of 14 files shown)


packages/editor/src/components/parts/program/editor/EditorPart.test.tsx
```diff
@@ -0,0 +1,62 @@
...
+    screen.debug();
+    expect(screen.getByTitle('There is no editor for this bean')).toBeInTheDocument();
+  });
+
+  test('full data', async () => {
+    renderPart({
+      javaClass: 'bla',
+      link: 'blabla',
+      userConfig: 'text'
+    });
```
ivy-lli, line 29: add test with meta data

packages/editor/src/components/parts/program/editor/EditorPart.tsx
```diff
@@ -0,0 +1,77 @@
+import { EditorData, Label, Script, Text, Widget } from '@axonivy/inscription-protocol';
+import { PartProps, usePartDirty, usePartState } from '../../../editors';
+import { useEditorContext, useMeta, useValidations } from '../../../../context';
+import { useEditorData } from './useEditorData';
+import { Input, MessageText, ScriptInput } from '../../../../components/widgets';
+
+export function useEditorPart(): PartProps {
+  const { config, defaultConfig, initConfig, reset } = useEditorData();
+  const compareData = (data: EditorData) => [data.userConfig, data.link];
```
ivy-lli, line 9: what is this link for?

packages/editor/src/components/parts/program/editor/EditorPart.tsx
```diff
@@ -0,0 +1,77 @@
+import { EditorData, Label, Script, Text, Widget } from '@axonivy/inscription-protocol';
+import { PartProps, usePartDirty, usePartState } from '../../../editors';
+import { useEditorContext, useMeta, useValidations } from '../../../../context';
+import { useEditorData } from './useEditorData';
+import { Input, MessageText, ScriptInput } from '../../../../components/widgets';
+
+export function useEditorPart(): PartProps {
+  const { config, defaultConfig, initConfig, reset } = useEditorData();
+  const compareData = (data: EditorData) => [data.userConfig, data.link];
+  const validation = useValidations(['javaClass']);
```
ivy-lli, line 10: I think you want to have the userconfig here right?

packages/editor/src/components/parts/program/editor/EditorPart.tsx
```diff
@@ -0,0 +1,77 @@
...
+        const sentences = message.split(/(?<=\.)\s+(?=\D|(?<!\.\s)\d)|\b(?=[A-Z]{8,}\b)/);
+        return (
+          <>
+            {sentences.map((sentence, index) => (
+              <div key={index}>
+                <MessageText message={{ severity: 'INFO', message: sentence.trim() }} />
+              </div>
+            ))}
+          </>
+        );
```
ivy-lli, line 54: let do css the job.
Why not simply write this message to the UI and let CSS do the linebreaks?
Because such a regex is just simply not maintainable...

packages/editor/src/components/parts/program/editor/EditorPart.tsx
```diff
@@ -0,0 +1,77 @@
...
+      } else {
+        return <MessageText message={{ severity: 'INFO', message: message }} />;
+      }
+    } else if (isScript(widget)) {
```
ivy-lli, line 58: if you do a `return` in an `if` statement you can simply do another `if` instead of an `else if`.
Because it sometimes not a good coding style to have an `else if` statement without a closing final else.

packages/editor/src/components/parts/program/editor/EditorPart.tsx
```diff
@@ -0,0 +1,77 @@
...
+    state,
+    reset: { dirty, action: () => reset() },
+    content: <EditorPart />
+  };
+}
+
+const EditorPart = () => {
+  const { config, update } = useEditorData();
+  const { context } = useEditorContext();
+  const editorItems = useMeta('meta/program/editor', { context, type: config.javaClass }, []);
```
ivy-lli, line 24: directly get the data here instead of everywhere below:
`const editorItems = useMeta('meta/program/editor', { context, type: config.javaClass }, []).data;`

packages/editor/src/components/parts/program/editor/EditorPart.test.tsx
ivy-lli: You can add meta data into the render method, like you can add data

ivy-lli (changes requested):
Looks promising, some change requests

packages/editor/src/components/parts/program/editor/EditorPart.tsx
ivy-edp: honestly i don't know exactly either. i have now removed the link.

packages/editor/src/components/parts/program/editor/EditorPart.test.tsx
```diff
@@ -9,24 +9,32 @@ const Part = () => {
...
+        }
+      }
     });
   }
 
-  test.only('empty data', async () => {
-    renderPart();
-    screen.debug();
+  test('empty data', async () => {
+    render(<Part />, { wrapperProps: { data: {} && { config: {} } } });
```
ivy-lli, line 28: I think you can remove the whole wrapperProps part

packages/editor/src/components/parts/program/editor/EditorPart.tsx
```diff
@@ -42,7 +42,7 @@ const EditorPart = () => {
       if (widget.multiline) {
         // Split the text using a regular expression that considers periods followed by spaces or numbers
         // and capital Words longer than 8 letters. But not when it's followed by space and a number
```
ivy-lli, line 44: remove this

packages/protocol/src/data/part-data.ts
```diff
@@ -63,4 +63,4 @@ export type ProgramInterfaceStartData = Pick<ConfigData, 'javaClass' | 'exceptio
 
 export type EventData = Pick<ConfigData, 'javaClass' | 'eventId' | 'timeout'>;
 
-export type EditorData = Pick<ConfigData, 'link' | 'userConfig' | 'javaClass'>;
```
ivy-lli: I think you can remove this link also from the useEditorData hook

ivy-lli (approved):
Some final comments, but I think good to merge in general 👍

ivy-lli:
An integration test for a java class with one input field would also be cool. but you can add one in a later PR if you like

packages/editor/src/components/parts/program/editor/Editor.css
```diff
@@ -0,0 +1,6 @@
+.message {
```
ivy-lli, line 1: please invent a new unique class name.
Css is normally not scoped, which means the CSS of the message widget now is active here (and vice versa), and if we change this one or the other it can also change the styles of the message widget...

ivy-lli (changes requested):

packages/editor/src/components/parts/program/editor/Editor.css
```diff
@@ -1,6 +1,9 @@
-.message {
+.infoTextArea {
```
ivy-lli, line 1: currently classes are all written with '-' -> so `info-text-area`
But as this is no infotext i would rename it to something like `info-text`

ivy-lli:
```css
.info-text {
  display: flex;
  gap: ...
  padding-inline: ...
  font-size: 12px;
}
.info-text p {
  margin: 0;
}```

File: packages/editor/src/components/parts/program/editor/EditorPart.tsx [hunk 12]
```diff
@@ -46,13 +46,13 @@ const EditorPart = () => {
           <>
             {sentences.map((sentence, index) => (
               <div key={index}>
```
ivy-lli, line 48: move this div out of iteration with `info-text` class

packages/editor/src/components/parts/program/editor/EditorPart.tsx
```diff
@@ -46,13 +46,13 @@ const EditorPart = () => {
...
-                <p className='message'>{sentence.trim()}</p>
+                <p className='infoTextArea'>{sentence.trim()}</p>
               </div>
             ))}
           </>
         );
       } else {
-        return <p className='message'>{message}</p>;
+        return <p className='infoTextArea'>{message}</p>;
```
ivy-lli, line 55: make this to a div

ivy-lli (changes requested):

ivy-lli (approved):

PR #315, opened 2023-10-10
Fix protocol gen on windows

ivy-lli:


ivy-rew (commented):
nice one; 
thanks for improving my first silly steps in writing js code :sweat_smile:

PR #317, opened 2023-10-11
XIVY-11798 Add CMS-Browser

ivy-edp:
First Version of CMS-Browser:
I would appreciate feedback for my current code.

ToDo:
- RequiredProjects: I don't quite understand what I need this for yet?
- Search: I don't know or still need to figure out how the search works in general and how I can additionally search for the content.
- Insert: If the current value is "", the result is still <%=%>. I have now called attention to this situation using the Helper Text, because I don't quite understand how to modify the accept().
- UT: I have written two simple UT, maybe I can further improve them
- IT: I have not written any IT's yet.

image

### DIFF
--- a/packages/editor/src/components/browser/Browser.css
+++ b/packages/editor/src/components/browser/Browser.css
@@ -39,6 +39,11 @@
   background-color: var(--input-bg);
   border: 1px solid var(--editor-bg);
   padding: var(--input-padding);
+  overflow: auto;
+  max-height: 50vh;
+  display: flex;
+  flex-direction: column;
+  gap: var(--size-1);
 }
 
 .dialog-title {
--- a/packages/editor/src/components/browser/CmsBrowser.tsx
+++ b/packages/editor/src/components/browser/CmsBrowser.tsx
@@ -1,15 +1,142 @@
-import { useState } from 'react';
-import { Input } from '../widgets';
+import { useMemo, useEffect, useState } from 'react';
+import { Checkbox, Collapsible, ExpandableCell, MessageText, SelectRow, Table, TableCell } from '../widgets';
 import { UseBrowserImplReturnValue } from './useBrowser';
+import { useEditorContext, useMeta } from '../../context';
+import {
+  ColumnDef,
+  ExpandedState,
+  RowSelectionState,
+  flexRender,
+  getCoreRowModel,
+  getExpandedRowModel,
+  getFilteredRowModel,
+  useReactTable
+} from '@tanstack/react-table';
+import { ContentObject } from '@axonivy/inscription-protocol';
+import { IvyIcons } from '@axonivy/editor-icons';
 
 export const CMS_BROWSER_ID = 'cms' as const;
 
 export const useCmsBrowser = (): UseBrowserImplReturnValue => {
-  const [value, setValue] = useState('cms');
+  const [value, setValue] = useState('');
 
-  return { id: CMS_BROWSER_ID, name: 'CMS', content: <CmsBrowser value={value} onChange={setValue} />, accept: () => value };
+  return {
+    id: CMS_BROWSER_ID,
+    name: 'CMS',
+    content: <CmsBrowser value={value} onChange={setValue} />,
+    accept: () => addIvyPathToValue(value)
+  };
 };
 
-const CmsBrowser = (props: { value: string; onChange: (value: string) => void }) => {
-  return <Input {...props} />;
+export const addIvyPathToValue = (value: string) => {
+  const fullPath = value.length === 0 ? '' : `ivy.cms.co("${value}")`;
+  return fullPath;
+};
+
+const CmsBrowser = ({ value, onChange }: { value: string; onChange: (value: string) => void }) => {
+  const { context } = useEditorContext();
+
+  const [requiredProject, setRequiredProject] = useState<boolean>(false);
+  const tree = useMeta('meta/cms/tree', { context, requiredProjects: requiredProject }, []).data;
+  const [selectedContentObject, setSelectedContentObject] = useState<ContentObject | undefined>();
+  const [showHelper, setShowHelper] = useState<boolean>(false);
+
+  const columns = useMemo<ColumnDef<ContentObject>[]>(
+    () => [
+      {
+        accessorFn: row => `${row.name} (${row.type})`,
+        id: 'name',
+        cell: cell => {
+          return (
+            <ExpandableCell
+              cell={cell}
+              title={cell.row.original.name}
+              icon={cell.row.original.type === 'FOLDER' ? IvyIcons.GoToSource : undefined}
+            />
+          );
+        }
+      }
+    ],
+    []
+  );
+
+  const [expanded, setExpanded] = useState<ExpandedState>(false as ExpandedState);
+  const [globalFilter, setGlobalFilter] = useState('');
+  const [rowSelection, setRowSelection] = useState<RowSelectionState>({});
+
+  const table = useReactTable({
+    data: tree,
+    columns: columns,
+    state: {
+      expanded,
+      globalFilter,
+      rowSelection
+    },
+    filterFromLeafRows: true,
+    enableRowSelection: true,
+    enableMultiRowSelection: false,
+    enableSubRowSelection: false,
+    onExpandedChange: setExpanded,
+    onGlobalFilterChange: setGlobalFilter,
+    onRowSelectionChange: setRowSelection,
+    getSubRows: row => row.children,
+    getCoreRowModel: getCoreRowModel(),
+    getExpandedRowModel: getExpandedRowModel(),
+    getFilteredRowModel: getFilteredRowModel()
+  });
+
+  useEffect(() => {
+    if (Object.keys(rowSelection).length !== 1) {
+      setSelectedContentObject({ name: '', children: [], fullPath: '', type: 'STRING', values: {} });
+      setShowHelper(false);
+      onChange('');
+      return;
+    }
+    const selectedRow = table.getRowModel().rowsById[Object.keys(rowSelection)[0]];
+    setSelectedContentObject(selectedRow.original);
+    setShowHelper(true);
+    onChange(selectedRow.original.fullPath);
+  }, [onChange, rowSelection, table]);
+
+  return (
+    <>
+      <Table search={{ value: globalFilter, onChange: setGlobalFilter }}>
+        <Checkbox label='Enable required Projects' value={requiredProject} onChange={() => setRequiredProject(!requiredProject)} />
+        <tbody>
+          {table.getRowModel().rows.map(row => (
+            <SelectRow key={row.id} row={row} isNotSelectable={row.original.type === 'FOLDER'}>
+              {row.getVisibleCells().map(cell => (
+                <TableCell key={cell.id}>{flexRender(cell.column.columnDef.cell, cell.getContext())}</TableCell>
+              ))}
+            </SelectRow>
+          ))}
+        </tbody>
+      </Table>
+
+      <Collapsible label='Helper Text' defaultOpen={showHelper}>
+        {value.length !== 0 && selectedContentObject ? (
+          <pre className='browser-helptext'>
+            <b>{value}</b>
+            <code>
+              {Object.entries(selectedContentObject.values).map(([key, value]) => (
+                <div key={key}>
+                  <b>{`${key}: `}</b>
+                  {value}
+                </div>
+              ))}
+            </code>
+          </pre>
+        ) : (
+          <pre className='browser-helptext'>
+            <MessageText
+              message={{
+                severity: 'INFO',
+                message: `No element selected.`
+              }}
+            />
+          </pre>
+        )}
+      </Collapsible>
+    </>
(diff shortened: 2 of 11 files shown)


packages/editor/src/components/browser/CmsBrowser.tsx
```diff
@@ -1,15 +1,151 @@
...
+      setSelectedContentObject({ name: '', children: [], fullPath: '', type: 'STRING', values: {} });
+      setShowHelper(false);
+      onChange('');
+      return;
+    }
+    const selectedRow = table.getRowModel().rowsById[Object.keys(rowSelection)[0]];
+    setSelectedContentObject(selectedRow.original);
+    setShowHelper(true);
+    onChange(selectedRow.original.fullPath);
+  }, [onChange, rowSelection, table]);
```
ivy-lli, line 100: instead of this useEffect, why not simply change the function on the onRowSelectionChange property?

packages/editor/src/components/browser/CmsBrowser.tsx
```diff
@@ -1,15 +1,151 @@
...
+const CmsBrowser = ({ value, onChange }: { value: string; onChange: (value: string) => void }) => {
+  const { context } = useEditorContext();
+  const tree = useMeta('meta/cms/tree', { context, requiredProjects: false }, []).data;
+  const [selectedContentObject, setSelectedContentObject] = useState<ContentObject>({
+    name: '',
+    children: [],
+    fullPath: '',
+    type: 'STRING',
+    values: {}
+  });
```
ivy-lli, line 44: instead of force an empty ContentObject, cann't you simply work with undefined? ContentObject if a selection is active, undefined if not.
`useState<ContentObject | undefined>()`

packages/editor/src/components/browser/CmsBrowser.tsx
```diff
@@ -1,15 +1,151 @@
...
+    onChange(selectedRow.original.fullPath);
+  }, [onChange, rowSelection, table]);
+
+  return (
+    <>
+      <Table search={{ value: globalFilter, onChange: setGlobalFilter }}>
+        <tbody>
+          {table.getRowModel().rows.map(row =>
+            row.original.type === 'FOLDER' ? (
+              <MessageRow key={row.id}>
```
ivy-lli, line 108: you don't display a message here right? So why not simply use a normal row? `<tr key={row.id}>`

ivy-lli: or even better improve the SelectRow by an attribute if the row is selectable or not. then you don't have to do this condition and duplication code here!

packages/editor/src/components/browser/CmsBrowser.tsx
```diff
@@ -1,15 +1,151 @@
...
+          </pre>
+        ) : (
+          <pre className='browser-helptext'>
+            <MessageText
+              message={{
+                severity: 'WARNING',
+                message: `No element selected; it simply returns '<%=%>'.`
+              }}
+            />
+          </pre>
```
ivy-lli, line 146: This should not be the case, so this warning is not true.

packages/editor/src/components/browser/CmsBrowser.tsx
```diff
@@ -1,15 +1,151 @@
...
+                ))}
+              </SelectRow>
+            )
+          )}
+        </tbody>
+      </Table>
+      <Collapsible label='Helper Text' defaultOpen={showHelper}>
+        {value.length !== 0 && selectedContentObject ? (
+          <pre className='browser-helptext'>
+            <p className='helperTitle'>{createFullPathCMS(value)}</p>
```
ivy-lli, line 126: There is no benefit from showing the `ivy.cms.co(...` stuff here. This is only some technical stuff and should only be added on the accept method! 
Simply show the full path here with the type, e.g. `/test/Bla (File)`

packages/editor/src/components/browser/CmsBrowser.tsx
```diff
@@ -1,15 +1,151 @@
...
+    id: CMS_BROWSER_ID,
+    name: 'CMS',
+    content: <CmsBrowser value={value} onChange={setValue} />,
+    accept: () => createFullPathCMS(value)
+  };
 };
 
-const CmsBrowser = (props: { value: string; onChange: (value: string) => void }) => {
-  return <Input {...props} />;
+export const createFullPathCMS = (value: string) => {
```
ivy-lli, line 30: this method does not calculate the full path, but add the ivy cms co. So rename it to what it does.

packages/editor/src/components/widgets/table/cell/ExpandableCell.tsx
```diff
@@ -22,13 +24,19 @@ export function ExpandableCell<TData>({ cell, isLoaded, loadChildren, isUnknown,
...
+            icon={IvyIcons.AngleDown}
+            className='row-expand-button'
+            aria-label={row.getIsExpanded() ? 'Collapse row' : 'Expand row'}
+            data-state={row.getIsExpanded() ? 'expanded' : 'collapsed'}
+            {...{ onClick: row.getToggleExpandedHandler() }}
+          />
+          {additionalIcon?.length == 0 && additionalIcon !== undefined && '🔵'}
+          {additionalIcon?.length !== 0 && additionalIcon !== undefined && (
+            <IvyIcon icon={IvyIcons[additionalIcon as keyof typeof IvyIcons]} />
+          )}
```
ivy-lli, line 38: No! You write typescript code and not javascript!
If your additionalIcon (which I would still rename to just icon) would be of type IvyIcons in the first place you didn't have to do all this ugly typechecking stuff!
Just render the icon if it is not undefined.

And I would not render this blue dot here, because it would not match with the other code. You already show the type of the content object right? I think this is enough for the start to give the user the hint if he can click on something or not. We can still improve this after we have more icons and restyle this tree tables

packages/editor/src/components/browser/Browser.css
```diff
@@ -67,6 +69,16 @@
...
+.dialog-content .browser-helptext p.helperTitle {
+  margin: 0 0 var(--size-2) 0;
+  font-weight: bold;
+  font-size: 12;
+}
+
+.dialog-content .browser-helptext .cmsValue {
+  margin-top: var(--size-1);
+}
+
```
ivy-lli, line 81: didn't I told you why it is not a good idea to use margins? If the p has default margins then set those to 0 or just don't use 'p'
Better add a display flex on the parent (.browser-helptext) with a gap. And the font-size is not needed on the title I think, as you already define it on the parent.

If you do so you don't need to do any stylings and you also don't need to give the value an own class, so you can simplify the html code a lot too. Se below.

packages/editor/src/components/browser/CmsBrowser.tsx
```diff
@@ -1,15 +1,151 @@
...
+          <pre className='browser-helptext'>
+            <p className='helperTitle'>{createFullPathCMS(value)}</p>
+            <code>
+              {Object.entries(selectedContentObject.values).map(([key, value]) => (
+                <div key={key} className='cmsValue'>
+                  <span className='key' style={{ fontWeight: 'bold' }}>
+                    {key}:{' '}
+                  </span>
+                  <span className='value'>{value}</span>
+                </div>
```
ivy-lli, line 134: I would simplify this code a lot... Just do something like
```<div key={key}><b>{`${key}: `}</b>{value}</div>```

**Review by ivy-lli — CHANGES_REQUESTED** [review]
Feedback

**Comment by ivy-lli** [comment]
> RequiredProjects: I don't quite understand what I need this for yet?

Why didn't you ask right away if you didn't understand what I was talking about? If your project has a dependency on another ivy project that has CMS stuff in it you can access this content too. Just add a dependent project and open the cms browser in an old swt mask.

> Search: I don't know or still need to figure out how the search works in general and how I can additionally search for the content.

I don't know this either, maybe it is not possible at all, didn't you read the documentation of react-table? https://tanstack.com/table/v8/

> Insert: If the current value is "", the result is still <%=%>. I have now called attention to this situation using the Helper Text, because I don't quite understand how to modify the accept().

Why? You just have to check if the value length is not 0 and only then apply the modifyAction. E.g. like this:
```ts
export const useModifyEditor = (modifyAction?: ModifyAction) => {
  const [editor, setEditor] = useState<monaco.editor.IStandaloneCodeEditor>();
  const modifyEditor = (value: string) => {
    if (!editor) {
      console.log('No editor set to modify');
      return;
    }
    const selection = editor.getSelection();
    if (!selection) {
      console.log('No selection found on editor');
      return;
    }
    const text = value.length > 0 && modifyAction ? modifyAction(value) : value;
  };
  return { setEditor, modifyEditor };
};
```

packages/editor/src/components/browser/CmsBrowser.tsx
ivy-lli: Use tree ``` for multiline code

ivy-edp:
> Why didn't you ask right away if you didn't understand what I was talking about? If your project has a dependency on another ivy project that has CMS stuff in it you can access this content too. Just add a dependent project and open the cms browser in an old swt mask.
I didn't ask directly because I wanted to try myself first if I see/understand what this is for and work through the other tasks in the CMS browser first. Sorry, I will ask directly next time. 
So in the CMS browser if "requiredProject" is true I have to add an additional button which allows to view the cms of the dependent project in an old swt mask? If yes, don't I need another entry in the inscription.ts like "openDependentCMS" (analog Java Class Selector add & edit)? Maybe we have to discuss this by phone
> I don't know this either, maybe it is not possible at all, didn't you read the documentation of react-table? https://tanstack.com/table/v8/
No I have not read the documentation yet, but I will do so and try to find out if it works.
> Why? You just have to check if the value length is not 0 and only then apply the modifyAction. E.g. like this:
Thank you very much. I haven't quite understood yet how this works with the method passes, so it was a bit difficult for me to understand where I have to put this useModifyEditor (in Browser.tsx or CmsBrowser.tsx) and how to call the CmsBrowser correctly afterwards (what do I have to pass). Instead of a renewed const useModifyEditor, can't I do the whole thing directly in useCmsBrowser and adjust the UseBrowserImplReturnValue? Maybe we also need to look at this in a phone call....
```
export const useCmsBrowser = (): UseBrowserImplReturnValue => {
  const [browserValue, setBrowserValue] = useState('');

  if (browserValue.length > 0) {
    return {
      id: CMS_BROWSER_ID,
      name: 'CMS',
      content: <CmsBrowser value={browserValue} onChange={setBrowserValue} />,
      accept: () => addIvyPathToValue(browserValue)
    };
  } else {
    return {
      id: CMS_BROWSER_ID,
      name: 'CMS',
      content: <CmsBrowser value={browserValue} onChange={setBrowserValue} />,
      accept: () => undefined
    };
  }
};
```

packages/editor/src/components/browser/CmsBrowser.tsx
ivy-lli: I think you need to call the setRowSelection too!

ivy-lli:
> So in the CMS browser if "requiredProject" is true I have to add an additional button which allows to view the cms of the dependent project in an old swt mask? If yes, don't I need another entry in the inscription.ts like "openDependentCMS" (analog Java Class Selector add & edit)? Maybe we have to discuss this by phone

No if you do a cms meta request with requiredproject true then data for all projects are delivered back.

packages/editor/src/components/widgets/table/row/SelectRow.tsx
```diff
@@ -2,8 +2,20 @@ import './SelectRow.css';
...
-  <tr className='selectable-row' data-state={row.getIsSelected() ? 'selected' : ''} onClick={row.getToggleSelectedHandler()}>
+export const SelectRow = <TData extends object>({
+  row,
+  children,
+  isNotSelectable
+}: {
+  row: Row<TData>;
+  children: ReactNode;
+  isNotSelectable?: boolean;
+}) => (
```
ivy-lli, line 13: as soon as it breaks all attributes below each other I start move the property type definition outside. Just a personal preference but I think it is much more readable:

```ts
type SelectRowProps<TData> = {
  row: Row<TData>;
  children: ReactNode;
  isNotSelectable?: boolean;
}

export const SelectRow = <TData extends object>({ row, children, isNotSelectable }: SelectRowProps<TData>) => (
```

ivy-lli (approved):
Cool. Thanks for the improvements!

packages/editor/src/components/parts/common/mapping-tree/MappingPart.test.tsx
ivy-edp, line 1: @ivy-lli This test has always failed. i had to adjust it. is this ok? i'm not sure why it always fails but could it be related to the fact that i added "icon" to the ExpandableCell?

ivy-lli: Yes, the expandablecell is use on other places too, so if you change it may have an effect on other locations too. It is ok, if it has no negative effect (looks stange or something the like)

ivy-lli (approved):
If I once give approval and you don't change something crucial again, you don't have to ask for approval again :)

PR #320, opened 2023-10-16
XIVY-11713 Fix form table

ivy-lli:
- Fix show correct properties from openapi spec
- Handle multiple same key


packages/editor/src/components/parts/rest/rest-request/rest-body/rest-parameter.test.ts
```diff
@@ -51,6 +71,13 @@ describe('RestParam', () => {
...
   test('to', () => {
-    expect(toMappings(params)).toEqual({ test: '123', petId: '4' });
+    expect(toRestMap(expressionParams)).toEqual({ file: ['123'], ownerId: ['4'] });
+  });
+
+  test('to - multiple same attributes', () => {
+    const sameParams = cloneObject(expressionParams);
+    sameParams.push(params[0]);
+    sameParams[2].expression = 'blabla';
+    expect(toRestMap(sameParams)).toEqual({ file: ['123', 'blabla'], ownerId: ['4'] });
```
ivy-rew, line 81: I'd consider to add this test with more information, why we support this very special mode.
so that later devs do have such a hard time with it

ivy-rew (approved):
:clap:

PR #324, opened 2023-10-17
XIVY-11798 Add improvements for all-browsertypes

ivy-edp:
Added string/file validation to cms browser (cms.co or cms.cr)
Added extended content-search to cms browser
Added EndPart - Input with Browser
Added EditableCell with Browser

### DIFF
--- a/packages/editor/src/components/browser/Browser.tsx
+++ b/packages/editor/src/components/browser/Browser.tsx
@@ -5,26 +5,27 @@ import { IvyIcons } from '@axonivy/editor-icons';
 import { useEditorContext } from '../../context';
 import { useState } from 'react';
 import { useAttributeBrowser } from './AttributeBrowser';
-import { useCmsBrowser } from './CmsBrowser';
 import { BrowserType, UseBrowserReturnValue } from './useBrowser';
 import { useFuncBrowser } from './FunctionBrowser';
 import { useDataTypeBrowser } from './DataTypeBrowser';
 import { useTableColBrowser } from './TableColBrowser';
 import { useSqlOpBrowser } from './SqlOperationBrowser';
 import { useCatPathChooserBrowser } from './CatPathChooser';
+import { CmsOptions, useCmsBrowser } from './CmsBrowser';
 
 type BrowserProps = UseBrowserReturnValue & {
   types: BrowserType[];
   accept: (value: string) => void;
   location: string;
+  cmsOptions?: CmsOptions;
 };
 
-const Browser = ({ open, onOpenChange, types, accept, location }: BrowserProps) => {
+const Browser = ({ open, onOpenChange, types, accept, location, cmsOptions }: BrowserProps) => {
   const { editorRef } = useEditorContext();
   const [active, setActive] = useState<BrowserType>(types[0]);
 
   const attrBrowser = useAttributeBrowser(location);
-  const cmsBrowser = useCmsBrowser();
+  const cmsBrowser = useCmsBrowser(cmsOptions);
   const funcBrowser = useFuncBrowser();
   const dataTypeBrowser = useDataTypeBrowser();
   const catPathChooserBrowser = useCatPathChooserBrowser();
--- a/packages/editor/src/components/browser/CmsBrowser.tsx
+++ b/packages/editor/src/components/browser/CmsBrowser.tsx
@@ -4,40 +4,52 @@ import { UseBrowserImplReturnValue } from './useBrowser';
 import { useEditorContext, useMeta } from '../../context';
 import {
   ColumnDef,
+  ColumnFiltersState,
   ExpandedState,
   RowSelectionState,
+  VisibilityState,
   flexRender,
   getCoreRowModel,
   getExpandedRowModel,
   getFilteredRowModel,
   useReactTable
 } from '@tanstack/react-table';
-import { ContentObject } from '@axonivy/inscription-protocol';
+import { ContentObject, ContentObjectType } from '@axonivy/inscription-protocol';
 import { IvyIcons } from '@axonivy/editor-icons';
 
 export const CMS_BROWSER_ID = 'cms' as const;
 
-export const useCmsBrowser = (): UseBrowserImplReturnValue => {
+export type CmsTypeFilter = 'STRING' | 'FILE' | 'NONE';
+
+export type CmsOptions = {
+  noApiCall?: boolean;
+  typeFilter?: CmsTypeFilter;
+};
+
+export const useCmsBrowser = (options?: CmsOptions): UseBrowserImplReturnValue => {
   const [value, setValue] = useState('');
 
   return {
     id: CMS_BROWSER_ID,
     name: 'CMS',
-    content: <CmsBrowser value={value} onChange={setValue} />,
-    accept: () => addIvyPathToValue(value)
+    content: <CmsBrowser value={value} onChange={setValue} noApiCall={options?.noApiCall} typeFilter={options?.typeFilter} />,
+    accept: () => value
   };
 };
 
-export const addIvyPathToValue = (value: string) => {
-  const fullPath = value.length === 0 ? '' : `ivy.cms.co("${value}")`;
-  return fullPath;
-};
+interface CmsBrowserProps {
+  value: string;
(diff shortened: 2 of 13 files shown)


packages/editor/src/components/widgets/table/cell/EditableCell.tsx
```diff
@@ -1,7 +1,11 @@
 import './EditableCell.css';
+import '../../code-editor/ScriptInput.css';
 import { CellContext, RowData } from '@tanstack/react-table';
 import { useEffect, useState } from 'react';
 import { Input } from '../../input';
+import { Browser, useBrowser } from '../../../../components/browser';
+import { usePath } from '../../../../context';
+import { useCodeEditorOnFocus } from '../../code-editor/useCodeEditor';
```
ivy-lli, line 8: looks like this hook no longer has the correct name nor location.
I think it should live in the browser folder and should be called useInputOnFocus or useOnFocus or useOnInputFocus or something the like

packages/editor/src/components/browser/Browser.tsx
```diff
@@ -17,14 +17,16 @@ type BrowserProps = UseBrowserReturnValue & {
   types: BrowserType[];
   accept: (value: string) => void;
   location: string;
+  shortPath?: boolean;
+  onlyFile?: boolean;
```
ivy-lli, line 21: I think this should be an own object as it is only for the cms browser, e.g.
`cmsOptions?: {shortPath: boolean, onlyFily: boolean}`

And the only usages currently always set both on true. Are really both needed?

ivy-lli (changes requested):
Good first step. Some comments

packages/editor/src/components/browser/Browser.tsx
ivy-edp: At the moment, not really. I could create a flag, let's say "shortPathOnlyFile." However, I've kept them separate in case there's a need to use them independently elsewhere, such as filtering by files or outputting only the "short" path. Should I reduce it to a single flag for the moment?

ivy-lli: So is there a need in future for both flags?
If not I would definitely reduce this to one flag, because you have now code that will never be run...
If yes, leave it as it is, but maybe move it to a cmsOptions object

ivy-edp: I have now adjusted the flags:
- I put them into an own object and changed the flag from "onlyFile" to "typeFilter" where it is now possible to filter by FILE, STRING or NONE. 
- I renamed the Flag "shortPath" to "noApiCall"

Additionally I extended the search, that it also searches the content and as soon as something is entered in the search, the whole tree is opened.

packages/editor/src/components/browser/Browser.tsx
```diff
@@ -5,26 +5,29 @@ import { IvyIcons } from '@axonivy/editor-icons';
...
 import { useCatPathChooserBrowser } from './CatPathChooser';
+import { useCmsBrowser } from './CmsBrowser';
+
+export type typeFilter = 'STRING' | 'FILE' | 'NONE';
 
 type BrowserProps = UseBrowserReturnValue & {
   types: BrowserType[];
   accept: (value: string) => void;
   location: string;
+  cmsOptions?: { noApiCall?: boolean; typFilter?: typeFilter };
```
ivy-lli, line 22: I would move this cmsOptions into the cms browser itself and only map it here

packages/editor/src/components/browser/CmsBrowser.tsx
```diff
@@ -4,40 +4,46 @@ import { UseBrowserImplReturnValue } from './useBrowser';
...
 } from '@tanstack/react-table';
-import { ContentObject } from '@axonivy/inscription-protocol';
+import { ContentObject, ContentObjectType } from '@axonivy/inscription-protocol';
 import { IvyIcons } from '@axonivy/editor-icons';
+import { typeFilter } from './Browser';
 
 export const CMS_BROWSER_ID = 'cms' as const;
 
-export const useCmsBrowser = (): UseBrowserImplReturnValue => {
+export const useCmsBrowser = (noApiCall?: boolean, typFilter?: typeFilter): UseBrowserImplReturnValue => {
```
ivy-lli, line 23: move the cms options in here!

```ts
export type CmsTypeFilter = 'STRING' | 'FILE' | 'NONE';

export type CmsOptions = {
  noApiCall?: boolean; 
  typFilter?: CmsTypeFilter
}

export const useCmsBrowser = (options?: CmsOptions): UseBrowserImplReturnValue => {
```

packages/editor/src/components/browser/Browser.tsx
ivy-lli: `cmsOptions?: CmsOptions`

packages/editor/src/components/browser/CmsBrowser.tsx
ivy-lli: or even future down to the CmsBrowser compoent itself.

ivy-lli: and "type" not "typ"...

packages/editor/src/components/browser/CmsBrowser.tsx
```diff
@@ -63,26 +79,44 @@ const CmsBrowser = ({ value, onChange }: { value: string; onChange: (value: stri
   const [expanded, setExpanded] = useState<ExpandedState>(false as ExpandedState);
   const [globalFilter, setGlobalFilter] = useState('');
   const [rowSelection, setRowSelection] = useState<RowSelectionState>({});
+  const [columnVisibility, setColumnVisibility] = useState<VisibilityState>({ type: false, values: false });
```
ivy-lli, line 82: totally fine. Not sure if you can simply set the visibility on the columndef and then maybe you don't even need this state here, but I'm not sure...

packages/editor/src/components/browser/CmsBrowser.tsx
```diff
@@ -63,26 +79,44 @@ const CmsBrowser = ({ value, onChange }: { value: string; onChange: (value: stri
...
+
+  const addIvyPathToValue = (value: string, type: ContentObjectType, noApiCall?: boolean) => {
+    if (value.length === 0) {
+      return '';
+    }
+    if (type === 'FILE') {
+      return noApiCall ? value : `ivy.cms.cr("${value}")`;
+    }
+    return noApiCall ? value : `ivy.cms.co("${value}")`;
+  };
```
ivy-lli, line 95: instead of check this noApiCall twice, make one check at the beginning:

```ts
const addIvyPathToValue = (value: string, type: ContentObjectType, noApiCall?: boolean) => {
  if (noApiCall || value.length === 0) {
    return value
  }
  if (type === 'FILE') {
    return `ivy.cms.cr("${value}")`;
  }
  return `ivy.cms.co("${value}")`;
};
```

packages/editor/src/components/widgets/input/InputWithBrowser.tsx
```diff
@@ -0,0 +1,27 @@
+import './Input.css';
+import { Browser, BrowserType, useBrowser } from '../../../components/browser';
+import { usePath } from '../../../context';
+import { ComponentProps, forwardRef } from 'react';
+import Input from './Input';
+import { typeFilter } from '../../../components/browser/Browser';
+
+export type InputProps = Omit<ComponentProps<'input'>, 'value' | 'onChange' | 'ref'> & {
```
ivy-lli, line 8: Better naming -> InputWithBrowserProps
And I think you don't use this type somewhere else, so you don't have to export it

ivy-lli (changes requested):

ivy-lli (approved):

packages/editor/src/components/browser/CmsBrowser.tsx
ivy-edp: I originally wanted to do it this way, but in v8 this is apparently no longer so easily possible via the columndef: https://github.com/TanStack/table/discussions/4106

From my understanding, the solution I have currently implemented (I believe) is intentional and has already been set up to allow for the faster integration of potential radio buttons for filtering.

PR #326, opened 2023-10-19
Update monaco to 0.44.0

ivy-lli:
- Update monaco-languageclient to 6.6.0
- Update vscode-ws-jsonrpc to 3.0.0
- Update monaco-editor/react to 4.6.0

### DIFF
--- a/packages/core/src/monaco-util.ts
+++ b/packages/core/src/monaco-util.ts
@@ -1,16 +1,3 @@
-import 'monaco-editor/esm/vs/editor/editor.all.js';
-
-// support all editor features
-import 'monaco-editor/esm/vs/editor/standalone/browser/inspectTokens/inspectTokens.js';
-import 'monaco-editor/esm/vs/editor/standalone/browser/iPadShowKeyboard/iPadShowKeyboard.js';
-import 'monaco-editor/esm/vs/editor/standalone/browser/quickAccess/standaloneCommandsQuickAccess.js';
-import 'monaco-editor/esm/vs/editor/standalone/browser/quickAccess/standaloneGotoLineQuickAccess.js';
-import 'monaco-editor/esm/vs/editor/standalone/browser/quickAccess/standaloneGotoSymbolQuickAccess.js';
-import 'monaco-editor/esm/vs/editor/standalone/browser/quickAccess/standaloneHelpQuickAccess.js';
-import 'monaco-editor/esm/vs/editor/standalone/browser/quickInput/standaloneQuickInputService.js';
-import 'monaco-editor/esm/vs/editor/standalone/browser/referenceSearch/standaloneReferenceSearch.js';
-import 'monaco-editor/esm/vs/editor/standalone/browser/toggleHighContrast/toggleHighContrast.js';
-
 // @ts-ignore
 // eslint-disable-next-line import/no-unresolved
 import editorWorker from 'monaco-editor/esm/vs/editor/editor.worker?worker';
--- a/packages/core/src/monaco-util.ts
+++ b/packages/core/src/monaco-util.ts
@@ -1,7 +1,6 @@
 import 'monaco-editor/esm/vs/editor/editor.all.js';
 
 // support all editor features
-import 'monaco-editor/esm/vs/editor/standalone/browser/accessibilityHelp/accessibilityHelp.js';
 import 'monaco-editor/esm/vs/editor/standalone/browser/inspectTokens/inspectTokens.js';
 import 'monaco-editor/esm/vs/editor/standalone/browser/iPadShowKeyboard/iPadShowKeyboard.js';
 import 'monaco-editor/esm/vs/editor/standalone/browser/quickAccess/standaloneCommandsQuickAccess.js';
@@ -16,17 +15,13 @@ import 'monaco-editor/esm/vs/editor/standalone/browser/toggleHighContrast/toggle
 // eslint-disable-next-line import/no-unresolved
 import editorWorker from 'monaco-editor/esm/vs/editor/editor.worker?worker';
 
-// eslint-disable-next-line import/no-unresolved
-import { StandaloneServices } from 'vscode/services';
-
-import { MonacoServices } from 'monaco-languageclient';
+import { initServices } from 'monaco-languageclient';
 
 import { buildWorkerDefinition } from 'monaco-editor-workers';
 
 export namespace MonacoUtil {
-  export function initStandalone() {
-    StandaloneServices.initialize({});
-    MonacoServices.install();
+  export async function initStandalone() {
+    await initServices();
     buildWorkerDefinition('../../node_modules/monaco-editor-workers/dist/workers', new URL('', window.location.href).href, false);
 
     self.MonacoEnvironment = {
--- a/integrations/standalone/src/index.tsx
+++ b/integrations/standalone/src/index.tsx
@@ -15,8 +15,8 @@ export async function start(): Promise<void> {
   const pid = URLParams.pid();
   const theme = URLParams.themeMode();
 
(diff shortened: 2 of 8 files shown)


ivy-lli:
Hi @martin-fleck-at 

I just make the update to the latest monaco work. 
But I'm still not sure if I did everything right...

Do I need to init some services on the new initService method? https://github.com/axonivy/inscription-client/pull/326/files#diff-cc3404152ba2eab35db7b48ebedf57de3d6c3e61fa851553f98ad710cd659a87R24

The accessabiltityHelp.js is no longer here to import, maybe included? https://github.com/axonivy/inscription-client/pull/326/files#diff-cc3404152ba2eab35db7b48ebedf57de3d6c3e61fa851553f98ad710cd659a87L4

packages/core/src/monaco-util.ts
```diff
@@ -1,7 +1,6 @@
 import 'monaco-editor/esm/vs/editor/editor.all.js';
 
 // support all editor features
-import 'monaco-editor/esm/vs/editor/standalone/browser/accessibilityHelp/accessibilityHelp.js';
```
martin-fleck-at, line 4: From what I can tell, all the imports were unified in a single import in https://github.com/TypeFox/monaco-languageclient/commit/2760e5d6324158421b7ac1bc37eaccbcd4f7a25b 
```
import 'monaco-editor/esm/vs/editor/edcore.main.js';
```

and then further adapted so I believe that now most of those services and features are used by default when you call `initServices` and we no longer have to explicitly add them.

martin-fleck-at (commented):
@ivy-lli For some reason I cannot properly build the standalone example because I always get the following error:
```
[vite]: Rollup failed to resolve import "monaco-editor/esm/vs/platform/product/common/productService.js" from "/home/martin/Git/Clients/AxonIvy/next-gen-inscription/node_modules/vscode/services.js".
This is most likely unintended because it can break your application at runtime.
If you do want to externalize this module explicitly add it to
`build.rollupOptions.external`
```
Have you seen that as well? Unfortunately, this prevents me from testing this change in more detail.

However, if we do not need any particular VS Code functionality for the Monaco editor, I also don't think we need to provide any custom override services. We could have a look at the theme service if that still gives us troubles but I'd only actually do this if we are missing something specific.

packages/core/src/monaco-util.ts
ivy-lli: Ah ok I need to check that, then I can maybe remove more of those here...

ivy-lli:
> @ivy-lli For some reason I cannot properly build the standalone example because I always get the following error:
> 
> ```
> [vite]: Rollup failed to resolve import "monaco-editor/esm/vs/platform/product/common/productService.js" from "/home/martin/Git/Clients/AxonIvy/next-gen-inscription/node_modules/vscode/services.js".
> This is most likely unintended because it can break your application at runtime.
> If you do want to externalize this module explicitly add it to
> `build.rollupOptions.external`
> ```
> 
> Have you seen that as well? Unfortunately, this prevents me from testing this change in more detail.

No I didn't have this issue...

> However, if we do not need any particular VS Code functionality for the Monaco editor, I also don't think we need to provide any custom override services. We could have a look at the theme service if that still gives us troubles but I'd only actually do this if we are missing something specific.

👍

ivy-lli:
Change seems to increase the size of the whole js bundle by about 2MB and the webjar by 0.55MB

ivy-cst (approved):
Looks good to me, thanks  :+1:

ivy-lli:
Hi @martin-fleck-at 
According to the size of the js bundle, I think it is connected with the change, that the monaco-languageclient no longer uses dynamic imports: https://github.com/TypeFox/monaco-languageclient#september-2023-v650
What do you think? 

Something other that I discovered, is that the monaco editor (with the update) no longer starts a separate worker for the editor. Before the update:
image

So I'm not sure if the code ... is needed any longer, but I'm also a little bit curious about removing it (and also that there is no longer a separate worker) because before there was a warning when the monaco couldn't start a worker:
image

Do you know something about this?

ivy-lli:
Seems to be because of the monaco-languageclient update...
Before:
image
After:
image

Checked with manualChunks for monaco and monaco-languageclient:
```
rollupOptions: {
   output: {
      manualChunks: {
         monaco: ['monaco-editor'],
         'monaco-languageclient': ['monaco-languageclient']
      }
   }
}
```

ivy-lli:
> Something other that I discovered, is that the monaco editor (with the update) no longer starts a separate worker for the editor. 

Seems that it is still started, but much later. Maybe only after some heavy load is detected or only for some specific services. So it seems ok to let this code as it is now.

martin-fleck-at:
> Hi @martin-fleck-at According to the size of the js bundle, I think it is connected with the change, that the monaco-languageclient no longer uses dynamic imports: https://github.com/TypeFox/monaco-languageclient#september-2023-v650 What do you think?

Ah, good catch! That is very likely the reason! Seems like the treeshaking approach is causing some problems with the usage of additional VS Code code and they are now treemending the monaco-editor again, resulting in a larger file. Especially the monaco-vscode-api already seems to be around 2.1MB according to https://github.com/CodinGame/monaco-vscode-api/issues/136 so there is probably not much we can do.

> So I'm not sure if the code ... is needed any longer, but I'm also a little bit curious about removing it (and also that there is no longer a separate worker) because before there was a warning when the monaco couldn't start a worker: image
> 
> Do you know something about this?
> 
> Seems that it is still started, but much later. Maybe only after some heavy load is detected or only for some specific services. So it seems ok to let this code as it is now.

Unfortunately I don't have any particular insights into this. Having a web worker is definitely a good but why it is now started later, I cannot say. However, I suggest to re-visit this if it becomes a problem, what do you think?

ivy-lli:
> Ah, good catch! That is very likely the reason! Seems like the treeshaking approach is causing some problems with the usage of additional VS Code code and they are now treemending the monaco-editor again, resulting in a larger file. Especially the monaco-vscode-api already seems to be around 2.1MB according to https://github.com/CodinGame/monaco-vscode-api/issues/136 so there is probably not much we can do.

Seems so, except don't use the monaco-languageclient... Do you have any experience with this case?

> Unfortunately I don't have any particular insights into this. Having a web worker is definitely a good but why it is now started later, I cannot say. However, I suggest to re-visit this if it becomes a problem, what do you think?

No idea why it is starting later, but as also no warning appears about blocking the main thread, I agree with you to check it again if it is a problem.

martin-fleck-at:
> > Ah, good catch! That is very likely the reason! Seems like the treeshaking approach is causing some problems with the usage of additional VS Code code and they are now treemending the monaco-editor again, resulting in a larger file. Especially the monaco-vscode-api already seems to be around 2.1MB according to [CodinGame/monaco-vscode-api#136](https://github.com/CodinGame/monaco-vscode-api/issues/136) so there is probably not much we can do.
> 
> Seems so, except don't use the monaco-languageclient... Do you have any experience with this case?

I don't have any direct experience with this. However, looking at the monaco-languageclient code, it seems to be an already thin small layer on top of the monaco-vscode-api, i.e., at least according to documentation we should only get the the following services: layout, environment, extension, files, quickAccess, languages, and model. If we pull in more, that is definitely something we should avoid or can clean after the fact (if it comes from the framework). However, from what I can see it is mainly the VS Code API, the Monaco editor and the translation that takes most of the space. However, since we need the monaco <-> vscode / LSP translation in our editor, I'm not sure how we can easily avoid that.

In the `node_modules` I see that a lot of stuff is checked I assume that rollup correctly only bundles up what we really use. But even if I only look at the compiled code, I'm still not sure how we reach 2.5MB to be honest.

ivy-lli:
Done with https://github.com/axonivy/inscription-client/pull/341
Updated to v0.44.0 even of the increased js size

PR #329, opened 2023-10-19
Add browser to rest signal

ivy-edp:
I have now created a combobox to which a browser can be added. I managed to implement this combobox in:
- Signal Boundary --> Signal Code
- Rest Client --> Request --> Body -->Entity type
- Rest Client --> Response --> Read body as type.

What is still missing in the Signal Boundary is the support of MacroInput. Currently there is an error message when an element is added to the combobox of the Signal Boundary via Attribute Browser. Should I also take a look at this in this PR or does the improvement of the signal input belong to another story?

### DIFF
--- a/packages/editor/src/components/parts/common/eventcode/EventCodeSelect.tsx
+++ b/packages/editor/src/components/parts/common/eventcode/EventCodeSelect.tsx
@@ -1,6 +1,7 @@
 import { EventCodeMeta } from '@axonivy/inscription-protocol';
 import { Combobox, ComboboxItem, FieldsetInputProps, IvyIcon } from '../../../widgets';
 import { IvyIcons } from '@axonivy/editor-icons';
+import ComboboxWithBrowser from '../../../../components/widgets/combobox/ComboboxWithBrowser';
 
 export type EventCodeItem = Pick<EventCodeMeta, 'eventCode'> & { info?: string } & ComboboxItem;
 
@@ -10,9 +11,10 @@ type EventCodeSelectProps = {
   eventCodes: EventCodeItem[];
   eventIcon: IvyIcons;
   comboboxInputProps?: FieldsetInputProps;
+  withBrowser?: boolean;
 };
 
-const EventCodeSelect = ({ eventCode, onChange, eventCodes, eventIcon, comboboxInputProps }: EventCodeSelectProps) => {
+const EventCodeSelect = ({ eventCode, onChange, eventCodes, eventIcon, comboboxInputProps, withBrowser }: EventCodeSelectProps) => {
   const comboboxItem = (item: EventCodeItem) => {
     return (
       <>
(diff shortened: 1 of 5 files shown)


ivy-lli:
The signal macro support is definitely not part of this story, but if you like you can have a look at it. But as now is cool stuff you can also do something else

packages/editor/src/components/widgets/combobox/ComboboxWithBrowser.tsx
```diff
@@ -0,0 +1,30 @@
+import './Combobox.css';
+import { ComponentProps } from 'react';
+import { Browser, BrowserType, useBrowser } from '../../../components/browser';
+import { usePath } from '../../../context';
+import Combobox, { ComboboxItem, ComboboxProps } from './Combobox';
+
+export type ComboboxWithBrowserProps<T extends ComboboxItem> = Omit<ComponentProps<'input'>, 'value' | 'onChange'> &
+  ComboboxProps<T> & {
+    browsers: BrowserType[];
+  };
```
ivy-lli, line 10: instead of copy past the comboboxprops, you could also simply extend it with the browser attribute

packages/editor/src/components/widgets/combobox/ComboboxWithBrowser.tsx
```diff
@@ -0,0 +1,30 @@
...
+import { Browser, BrowserType, useBrowser } from '../../../components/browser';
+import { usePath } from '../../../context';
+import Combobox, { ComboboxItem, ComboboxProps } from './Combobox';
+
+export type ComboboxWithBrowserProps<T extends ComboboxItem> = Omit<ComponentProps<'input'>, 'value' | 'onChange'> &
+  ComboboxProps<T> & {
+    browsers: BrowserType[];
+  };
+
+const ComboboxInputWithBrowser = <T extends ComboboxItem>({
```
ivy-lli, line 12: I would rename this to ComboboxWithBrowser

packages/editor/src/components/parts/common/eventcode/EventCodeSelect.tsx
```diff
@@ -10,9 +11,10 @@ type EventCodeSelectProps = {
   eventCodes: EventCodeItem[];
   eventIcon: IvyIcons;
   comboboxInputProps?: FieldsetInputProps;
+  withBrowser?: boolean;
```
ivy-lli, line 14: instead of this flag, I could also imagine to just pipe the types as parameter and if it is undefined or an empty array just don't render the browser.
but this solution is also fine.

packages/editor/src/components/parts/common/eventcode/EventCodeSelect.tsx
```diff
@@ -29,7 +31,18 @@ const EventCodeSelect = ({ eventCode, onChange, eventCodes, eventIcon, comboboxI
     );
   };
 
-  return <Combobox value={eventCode} onChange={onChange} items={eventCodes} comboboxItem={comboboxItem} {...comboboxInputProps} />;
+  return withBrowser === false || withBrowser === undefined ? (
```
ivy-lli, line 34: withBrowser = undefined and withBrowser = false results in the same, so just turn it to make it simpler:

```ts
return withBrowser ? (
    <ComboboxInputWithBrowser
      browsers={['attr']}
      value={eventCode}
      onChange={onChange}
      items={eventCodes}
      comboboxItem={comboboxItem}
      {...comboboxInputProps}
    />
  ) : (
    <Combobox value={eventCode} onChange={onChange} items={eventCodes} comboboxItem={comboboxItem} {...comboboxInputProps} />
  );
```

ivy-lli (approved):

PR #330, opened 2023-10-20
Added Combobox with Browser and Macro func.

ivy-edp:
I have successfully implemented a version of the combobox with integrated browser and macro functionality. I achieved this by modifying the existing combobox component and managing its behavior using flags. I hope this approach is acceptable, or do you think it would have been better to keep them separate?

As a result, I was able to eliminate the previously created component, ComboboxWithBrowser, as its functionality is now controlled through the "BrowserTypes" flag.

ToDo: Adjust IT for the Signal Boundary because it is a Macrofield now. And maybe some additional tests for the new macro-functionality in the signal boundary?


packages/editor/src/components/widgets/combobox/Combobox.tsx
```diff
@@ -66,18 +82,45 @@ const Combobox = <T extends ComboboxItem>({ items, itemFilter, comboboxItem, val
...
+  const { setEditor, modifyEditor } = useModifyEditor(value => `<%=${value}%>`);
+  const path = usePath();
+  const browser = useBrowser();
+  const { isFocusWithin, focusValue, focusWithinProps } = useOnFocus(value, onChange);
 
   return (
     <div className='combobox'>
-      <div className='combobox-input'>
-        <input className='input' {...getInputProps()} {...inputProps} disabled={readonly} />
+      <div className='combobox-input' {...(macro !== undefined ? { ...focusWithinProps, tabIndex: 1 } : {})}>
```
ivy-lli, line 92: useless code

`{...(macro ? { ...focusWithinProps, tabIndex: 1 } : {})}`

packages/editor/src/components/widgets/combobox/Combobox.tsx
```diff
@@ -66,18 +82,45 @@ const Combobox = <T extends ComboboxItem>({ items, itemFilter, comboboxItem, val
...
+        {macro === true ? (
```
ivy-lli, line 93: useless code
`{macro ? (`

ivy-lli (approved):
A little bit complicated code, but I don't have a better idea at the moment. so go for it

ivy-edp:
Perfect thanks 👍
@ivy-lli I had to adjust the IT of the signal code, because the locator with the integration of the SingleLineCodeEditor was no longer set correctly. After spending some time attempting to modify the combobox pageobject to address this issue without success (because I probably have not quite understood it), I temporarily modified the IT itself, changing the .combobox to a .macroInput, which resolved the problem.

However, I think it would be nicer to have a new pageobject called MacroCombobox? I could try this improvement in another PR. What do you think?

ivy-lli:
Makes sense.
IMO it is fine as it is now. As long as we don't have a second MacroCombobox it should be enough to test only the macro part.

PR #332, opened 2023-10-25
XIVY-12548 Add Datatype Browser

ivy-edp:
First version of the datatype browser and unittests.
I commented out one useEffect in the Datatype-Browser (see reason in the comment itself). What do you think? Should this filter be handled in the backend instead?

Also I used margin in the Browser.css at one point because I couldn't (quickly) achieve it with display: flex and gap. I might try that again to improve it.

Additionally I added a little help, if someone has the checkbox "Use Type as List" set and then searches for lists, he gets a corresponding info that he has to uncheck it if he wants to see lists. Maybe unnecessary (I can also remove it) but I wanted to try it out :-).

ToDo (for another PR):
- IT for Mock
- Still missing the fixed metas for the Ivy datatypes
- Due to the lazy search it is not possible to find elements that are not in the dataClasses by single letters (e.g. all elements with name "List" can only be found if "List" is entered in the search. These do not appear if e.g. only "Li" was entered). Is this ok?
- In the old datatype browser is a separator "Workspace-matches" do I need to include something like this?
image

### DIFF
--- a/packages/editor/src/components/browser/DataTypeBrowser.test.tsx
+++ b/packages/editor/src/components/browser/DataTypeBrowser.test.tsx
@@ -55,9 +55,8 @@ describe('DataTypeBrowser', () => {
 
   test('renderCombinedDatatype', async () => {
     renderBrowser();
-    await userEvent.click(await screen.findByPlaceholderText('Search'));
+    await userEvent.click(screen.getByPlaceholderText('Search'));
     await userEvent.keyboard('Person');
-    screen.debug();
     await TableUtil.assertRowCount(1);
   });
 
@@ -65,7 +64,7 @@ describe('DataTypeBrowser', () => {
     let data = '';
     renderBrowser({ accept: value => (data = value) });
     await userEvent.click(await screen.findByRole('cell', { name: 'AddContactData' }));
-    await userEvent.click(screen.getByTestId('accept'));
+    await userEvent.click(screen.getByRole('button', { name: '' }));
     expect(data).toEqual('ch.ivyteam.documentation.project.AddContactData');
   });
 
@@ -74,7 +73,7 @@ describe('DataTypeBrowser', () => {
     renderBrowser({ accept: value => (data = value) });
     await userEvent.click(await screen.findByRole('cell', { name: 'AddContactData' }));
     await userEvent.click(await screen.findByRole('cell', { name: 'AddContactData' }));
-    await userEvent.click(screen.getByTestId('accept'));
+    await userEvent.click(screen.getByRole('button', { name: '' }));
     expect(data).toEqual('');
     await userEvent.click(await screen.findByText('Helper Text'));
     expect(screen.getByTitle(`No element selected.`)).toBeInTheDocument();
@@ -85,19 +84,7 @@ describe('DataTypeBrowser', () => {
     renderBrowser({ accept: value => (data = value) });
     await userEvent.click(await screen.findByRole('cell', { name: 'AddContactData' }));
     await userEvent.click(await screen.findByText('Use Type as List'));
-    await userEvent.click(screen.getByTestId('accept'));
-    expect(data).toEqual('java.util.List<AddContactData>');
-  });
-
-  test('listGenericWithList', async () => {
-    renderBrowser();
-    await userEvent.click(await screen.findByRole('cell', { name: 'AddContactData' }));
-    await userEvent.click(await screen.findByText('Use Type as List'));
-    await userEvent.click(await screen.findByRole('cell', { name: 'AddContactData' }));
-    await userEvent.click(await screen.findByPlaceholderText('Search'));
-    await userEvent.keyboard('List');
-    expect(screen.getByTitle('The "Use Type as List" checkmark is still enabled, causing lists not to be displayed.')).toBeInTheDocument();
-    await userEvent.click(await screen.findByText('Use Type as List'));
-    await TableUtil.assertRowCount(1);
+    await userEvent.click(screen.getByRole('button', { name: '' }));
+    expect(data).toEqual('java.util.List<ch.ivyteam.documentation.project.AddContactData>');
   });
 });
--- a/packages/editor/src/components/browser/DataTypeBrowser.test.tsx
+++ b/packages/editor/src/components/browser/DataTypeBrowser.test.tsx
@@ -1,0 +1,103 @@
+import { TableUtil, render, screen, userEvent } from 'test-utils';
+import { useDataTypeBrowser } from './DataTypeBrowser';
+
+const Browser = (props: { location: string; accept: (value: string) => void }) => {
+  const browser = useDataTypeBrowser();
+  return (
+    <>
+      {browser.content}
+      <button data-testid='accept' onClick={() => props.accept(browser.accept())} />
+    </>
+  );
+};
+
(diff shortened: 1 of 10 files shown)


ivy-lli:
> Still missing the fixed metas for the Ivy datatypes

rew task

> Due to the lazy search it is not possible to find elements that are not in the dataClasses by single letters (e.g. all elements with name "List" can only be found if "List" is entered in the search. These do not appear if e.g. only "Li" was entered). Is this ok?

sounds as a rew task too

> In the old datatype browser is a separator "Workspace-matches" do I need to include something like this?

I would ignore this for a start...

packages/editor/src/components/widgets/collapsible/Collapsible.tsx
```diff
@@ -5,15 +5,21 @@ import { IvyIcons } from '@axonivy/editor-icons';
...
-    if (defaultOpen) {
-      setOpen(defaultOpen);
+    if (autoClosable) {
+      if (defaultOpen !== undefined) {
+        setOpen(defaultOpen);
+      }
+    } else {
+      if (defaultOpen) {
+        setOpen(defaultOpen);
+      }
```
ivy-lli, line 20: so is this required, that this stays open?
If not, I would definitely vote for keeping it simple and just close it...

packages/editor/src/components/browser/DataTypeBrowser.tsx
```diff
@@ -1,13 +1,225 @@
...
-  return <Input {...props} />;
+  const { context } = useEditorContext();
+
+  const [datatypeList, setDatatypeList] = useState<JavaType[]>([]);
+  const [mainFilter, setMainFilter] = useState<string>('');
+  const allDatatypes: JavaType[] = useMeta('meta/scripting/allTypes', { context, type: mainFilter }, []).data;
+  const dataClasses: DataClass[] = useMeta('meta/scripting/dataClasses', context, []).data;
+
+  const [typeAsList, setTypeAsList] = useState(false);
+  const [listtypeSelected, setListtypeSelected] = useState(false);
```
ivy-lli, line 37: -> listTypeSelected, -> setListTypeSelected

packages/editor/src/components/browser/DataTypeBrowser.tsx
```diff
@@ -1,13 +1,225 @@
...
+    onGlobalFilterChange: setGlobalFilter,
+    onRowSelectionChange: setRowSelection,
+    getCoreRowModel: getCoreRowModel(),
+    getExpandedRowModel: getExpandedRowModel(),
+    getFilteredRowModel: getFilteredRowModel()
+  });
+
+  const addListGeneric = (value: JavaType, typeAsList: boolean) => {
+    if (typeAsList) {
+      return 'java.util.List<' + value.simpleName + '>';
```
ivy-lli, line 138: does the simple name work in every usecase?

ivy-lli: or better is this the same behavior like the old browser for every type

packages/editor/src/components/browser/DataTypeBrowser.tsx
```diff
@@ -1,13 +1,225 @@
...
+                    severity: 'WARNING',
+                    message: `The "Use Type as List" checkmark is still enabled, causing lists not to be displayed.`
+                  }}
+                />
+              )}
+              <Checkbox label='Use Type as List' value={typeAsList} onChange={() => setTypeAsList(!typeAsList)} />
+            </>
+          ) : null}
+        </>
+      )}
```
ivy-lli, line 200: also here, I would keep it simple.
I would just filter the list type from the possible types (maybe even in the backend!)

packages/protocol/src/data/inscription.ts
ivy-lli, line 1: you should not safe this file manually.
Technically this should be ingored by prettier, but it seems not to work anymore...

ivy-lli (commented):
Good start 👍

packages/editor/src/components/browser/DataTypeBrowser.tsx
```diff
@@ -1,13 +1,225 @@
...
+  const allDatatypes: JavaType[] = useMeta('meta/scripting/allTypes', { context, type: mainFilter }, []).data;
+  const dataClasses: DataClass[] = useMeta('meta/scripting/dataClasses', context, []).data;
+
+  const [typeAsList, setTypeAsList] = useState(false);
+  const [listtypeSelected, setListtypeSelected] = useState(false);
+
+  const [selectedDataType, setSelectedDataType] = useState<JavaType | undefined>();
+  const [showHelper, setShowHelper] = useState(false);
+
+  // Shorter useEffect for merging useMeta's (unfortunately i noticed that in the megat "allTypes" there are also types from the meta "dataClasses", so i created a new useEffect which removes these duplicates.)
```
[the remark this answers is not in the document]
ivy-edp: You're right, I've checked, and all types in 'dataClasses' are also present in 'allTypes'. Consequently, I could actually do away with this filter and just return 'allTypes' as soon as a search is initiated. However, the consequence of this is that if someone searches for a data class, the search is somewhat slower... I'm still looking into it with Reguel.

packages/editor/src/components/browser/DataTypeBrowser.test.tsx
```diff
@@ -0,0 +1,103 @@
...
+    await userEvent.keyboard('Person');
+    screen.debug();
+    await TableUtil.assertRowCount(1);
+  });
+
+  test('accept', async () => {
+    let data = '';
+    renderBrowser({ accept: value => (data = value) });
+    await userEvent.click(await screen.findByRole('cell', { name: 'AddContactData' }));
+    await userEvent.click(screen.getByTestId('accept'));
```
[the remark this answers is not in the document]
ivy-edp: I used 'testid' because it is already used in the testing of the Attribute Browser and the CMS Browser. But I can change it if you prefer.

packages/editor/src/components/widgets/collapsible/Collapsible.tsx
ivy-edp: I've adjusted the component with the 'autoClosable' feature so that the Collapsible continues to work the same as before everywhere, and in special cases like the browsers, the open/close function can be ensured. However, I believe we'll discuss this issue briefly tomorrow, and then we'll have clarity on how the helper text should behave. For now, I think I'll leave it like this, and if it proves to be too much, I can revert to the simpler version without the automatic closing support.

packages/editor/src/components/browser/DataTypeBrowser.tsx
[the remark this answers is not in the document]
ivy-edp: I discussed it with Reguel. In theory, I could do without this filter (since everything is covered in 'allTypes'), but I would still prefer to keep this filter in the frontend. This speeds up the search and provides immediate feedback as you type your search query. I hope this is acceptable to you as well.

packages/editor/src/components/browser/DataTypeBrowser.test.tsx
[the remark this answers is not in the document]
ivy-lli: I would only use the testId if nothing else is working because the testid will be included in the final build product, this means test code inside of the product. And if there is no big reason for this, this should be avoided :)

PR #335, opened 2023-10-27
XIVY-12548 Add browsertype to accept

ivy-edp:
I could adjust the 'accept' to include the browser type, and based on that, along with a flag to ensure that it only affects the ScriptArea, I was able to achieve the "import functionality." 

Unfortunately, all the tests of the components with a combobox are failing. As you can see, I haven't made any changes to the combobox or anything similar, so I have no idea why this is suddenly failing, especially when it didn't fail in the PR where I added the editor to the combobox. What should I do now? Should I revert the macro functionality of the combobox, or should I mock the whole thing (as you briefly mentioned before)? If the mock approach is appropriate, we should have a quick call again, as I didn't quite understand how that works.

### DIFF
--- a/packages/editor/src/components/widgets/code-editor/ScriptArea.tsx
+++ b/packages/editor/src/components/widgets/code-editor/ScriptArea.tsx
@@ -6,7 +6,7 @@ import { usePath } from '../../../context';
 
 const ScriptArea = (props: CodeEditorAreaProps) => {
   const browser = useBrowser();
-  const { setEditor, modifyEditor } = useModifyEditor();
+  const { setEditor, modifyEditor } = useModifyEditor({ scriptAreaDatatypeEditor: true });
   const path = usePath();
 
   return (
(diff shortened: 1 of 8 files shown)


ivy-edp:
Okay, never mind, I just found out why the error occurs. To make sure in the useCodeEditor that the line "Import XY" always starts on the first line of the editor, I currently make "range: new monaco.Range(1, 1, 1, 1)" and I think here is the problem

packages/editor/src/components/widgets/code-editor/ScriptArea.tsx
```diff
@@ -6,7 +6,7 @@ import { usePath } from '../../../context';
 
 const ScriptArea = (props: CodeEditorAreaProps) => {
   const browser = useBrowser();
-  const { setEditor, modifyEditor } = useModifyEditor();
+  const { setEditor, modifyEditor } = useModifyEditor(undefined, true);
```
ivy-lli, line 9: normally I use an object if I have multiple optional attributes, so you only need to define the once you need.
e.g.:
```ts
export const useModifyEditor = (options?: {modifyAction?: ModifyAction, scriptAreaDatatypeEditor?: boolean})

useModifyEditor({scriptAreaDatatypeEditor: true});
```

ivy-lli (approved):
Looks good

ivy-lli (approved):
🆒

PR #336, opened 2023-10-31
New design inscription view

ivy-edp:


integrations/standalone/tests/screenshots/generic-tab.spec.ts
ivy-lli, line 1: enable all this test again?

packages/editor/src/App.css
```diff
@@ -1,75 +1,85 @@
...
+  --N600: #a3a3a3;
+  --N700: #858585;
+  --N800: #757575;
+  --N900: #4a4a4a;
 
-  --accordion-header-bg: #ffffff;
-  --accordion-bg: #fcfcfc;
+  --error-color: #ee4a52;
+  --warning-color: #ffac5e;
+  --success-color: #47c46b;
```
ivy-lli, line 20: rename this to something more matching?

packages/editor/src/components/browser/Browser.css
```diff
@@ -1,30 +1,31 @@
 .dialog-overlay {
-  background-color: rgba(0, 0, 0, 0.44);
   position: fixed;
   inset: 0;
-  animation: overlayShow 150ms cubic-bezier(0.16, 1, 0.3, 1);
+  /* background-color: rgba(0, 0, 0, 0.44); */
```
ivy-lli, line 4: remove comment?

packages/editor/src/components/editors/part/Part.tsx
```diff
@@ -8,16 +8,20 @@ import { PartProps } from './usePart';
...
           control={part.reset.dirty ? { icon: IvyIcons.Undo, label: `Reset ${part.name}`, action: part.reset.action } : undefined}
           title={part.state.validations.map(val => val.message).join('\n')}
+          isAccordionOpen={isAccordionOpen}
         >
-          <span className='accordion-state' data-state={part.state.state} data-dirty={part.reset.dirty} />
-          {part.name}
+          <div className='part-header-left'>{part.name}</div>
+          <div className='part-header-right'>
+            <span className='accordion-state' data-state={part.state.state} data-dirty={part.reset.dirty} />
+          </div>
```
ivy-lli, line 24: why is this needed?

packages/editor/src/components/widgets/accordion/Accordion.tsx
```diff
@@ -32,15 +32,18 @@ type AccordionHeaderProps = {
...
+export const AccordionHeader = ({ title, control, children, isAccordionOpen }: AccordionHeaderProps) => (
   <Header className='accordion-header' title={title}>
-    <Trigger className='accordion-trigger'>{children}</Trigger>
     <div className='accordion-header-group'>
-      {control && <Button icon={control.icon} onClick={control.action} aria-label={control.label} />}
       <IvyIcon icon={IvyIcons.AngleDown} />
     </div>
+    <Trigger className='accordion-trigger'>{children}</Trigger>
+    {isAccordionOpen && control && (
+      <Button icon={control.icon} onClick={control.action} aria-label={control.label} style={{ paddingRight: 'var(--size-3)' }} />
```
ivy-lli, line 45: why not simply a general left right padding on the whole accoridon header?

packages/editor/src/components/widgets/accordion/Accordion.tsx
```diff
@@ -32,15 +32,18 @@ type AccordionHeaderProps = {
   title?: string;
   control?: { label: string; icon: IvyIcons; action: () => void };
   children: ReactNode;
+  isAccordionOpen: boolean;
 };
 
-export const AccordionHeader = ({ title, control, children }: AccordionHeaderProps) => (
+export const AccordionHeader = ({ title, control, children, isAccordionOpen }: AccordionHeaderProps) => (
```
ivy-lli, line 38: we are in the accordion, so imo no needed to duplicate this here -> rename to isOpen

packages/editor/src/components/widgets/code-editor/CodeEditor.css
```diff
@@ -1,10 +1,10 @@
 .code-editor .code-input {
   border-radius: var(--border-radius);
-  border: var(--input-border);
-  background: var(--input-bg);
+  border: var(--basic-border) !important;
```
ivy-lli, line 3: !important is mostly not a good idea... 
Why is this needed here?

packages/editor/src/components/widgets/combobox/Combobox.css
```diff
@@ -27,21 +27,31 @@
...
-.combobox-menu-entry:where(.hover, .selected) {
-  background: var(--editor-bg);
+.combobox-menu-entry:last-child {
+  border-bottom: none;
+}
+.combobox-menu-entry:where(.hover) {
+  background: var(--N50);
+}
+.combobox-menu-entry:where(.selected) {
+  background: var(--N100);
```
ivy-lli, line 54: if you only have one "where clause", the where makes no sense, you can simply write `.combobox-menu-entry.hover`

The where is cool if you need to apply the same styles to multiple different things, e.g. 
```css
.combobox-menu-entry:where(.hover, .selected) {
  background: var(--editor-bg);
}

/*instead of */
.combobox-menu-entry.hover, 
.combobox-menu-entry.selected {
  background: var(--editor-bg);
}
```

packages/editor/src/components/widgets/fieldset/Fieldset.css
```diff
@@ -25,10 +25,10 @@
   gap: var(--size-2);
 }
 .fieldset-input.fieldset-error {
-  --input-border: 1px solid var(--error-color);
+  --basic-border: 1px solid var(--error-color);
```
ivy-lli, line 28: be careful here. 
Everything inside this which uses this variable will have a red or yellow border

packages/editor/src/components/widgets/popover/Popover.css
```diff
@@ -1,13 +1,18 @@
 .popover-content {
   border-radius: var(--border-radius);
-  padding: var(--size-3);
   width: min(90vw, 400px);
-  background-color: var(--editor-bg);
-  box-shadow: var(--editor-shadow);
+  background-color: var(--N25);
+  /* box-shadow: var(--editor-shadow); */
```
ivy-lli, line 5: remove?

packages/editor/src/components/widgets/select/Select.css
```diff
@@ -32,20 +32,31 @@
...
+  border-bottom: none; /* Remove the border from the last entry */
+}
+.select-menu-entry:where(.hover) {
+  background: var(--N50);
 }
-.select-menu-entry:where(.hover, .selected) {
-  background: var(--editor-bg);
+
+.select-menu-entry:where(.selected) {
+  background: var(--N100);
```
ivy-lli, line 61: remove where

packages/editor/src/App.css
```diff
@@ -1,75 +1,85 @@
...
+  --transparent: rgba(0, 0, 0, 0);
 
-  --seperator-color: #7a7a7a;
-  --tab-inactive-border: #a0a0a0;
+  /* ------borders------*/
+  --basic-border: 0.75px solid var(--N200);
+  --thick-border: 1.25px solid var(--N200);
```
ivy-lli, line 25: be careful with comma px... makes maybe sense in figma but not in the browser with a normal zoom and no 4k monitor...

packages/editor/src/App.css
ivy-edp: For example? I actually find the names quite fitting, don't you? The "error-color" is only used for error messages, "warning" only for warnings, and "success" only for successful cases. Otherwise, could I simply omit the "-color"?

ivy-lli: Also true :)

packages/editor/src/components/editors/part/Part.tsx
ivy-edp: I used this to ensure that the Accordion State is displayed aligned to the right.

packages/editor/src/components/widgets/code-editor/CodeEditor.css
ivy-edp: That's a good question; I don't know why I set this. I believe it's because it wasn't inheriting the border, but I removed it, checked, and at first glance, everything seems to be working fine even without "!important.

packages/editor/src/components/editors/part/Part.tsx
ivy-lli: I mean the part-header left and right class. Isn't it just possible to do this stuff with flex and justify-content: space-between?

packages/editor/src/App.css
ivy-edp: Yes, you are right, i changed it to 1px and 2px for now 👍

integrations/standalone/tests/screenshots/generic-tab.spec.ts
ivy-edp: I disabled it because the tests were no longer functioning. For some reason, the GENERIC_PID is no longer correct. When running the tests, only the screen "Please select a process element" appears.

PR #338, opened 2023-11-02
Cms browser add string button

ivy-edp:
After clicking the Add button, no window opens. Currently I do not see what is wrong. when I look at the code in the core I do not think I need to add anything else.
image

### DIFF
--- a/packages/editor/src/components/browser/CmsBrowser.tsx
+++ b/packages/editor/src/components/browser/CmsBrowser.tsx
@@ -1,7 +1,7 @@
 import { useMemo, useEffect, useState } from 'react';
-import { Checkbox, Collapsible, ExpandableCell, MessageText, SelectRow, Table, TableCell } from '../widgets';
+import { ActionCell, Checkbox, Collapsible, ExpandableCell, MessageText, SelectRow, Table, TableCell } from '../widgets';
 import { UseBrowserImplReturnValue } from './useBrowser';
-import { useEditorContext, useMeta } from '../../context';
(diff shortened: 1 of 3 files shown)


packages/editor/src/components/browser/CmsBrowser.tsx
```diff
@@ -140,7 +142,10 @@ const CmsBrowser = ({ value, onChange, noApiCall, typeFilter }: CmsBrowserProps)
 
   return (
     <>
-      <Checkbox label='Enable required Projects' value={requiredProject} onChange={() => setRequiredProject(!requiredProject)} />
+      <div className='cms-add-header'>
+        <Checkbox label='Enable required Projects' value={requiredProject} onChange={() => setRequiredProject(!requiredProject)} />
+        <Button icon={IvyIcons.Add} aria-label={'Create new CMS-String'} title={'Create new CMS-String'} onClick={() => newAction()} />
```
ivy-lli, line 147: Did you debug the backend code?
I think you need to add here some payload to the action, right @ivy-rew?

ivy-rew: yes; I accept the 'uri:String' of the currently selected element as parent container

ivy-rew:
> After clicking the Add button, no window opens. Currently I do not see what is wrong. when I look at the code in the core I do not think I need to add anything else. 

sorry, since it's pure UI/SWT code not accessible anywhere else I couldn't write a test for it either. So quit possible that it doesn't really work.  ... we could debug it together, once you have a client calling that stuff :)

PR #339, opened 2023-11-02
XIVY-12552 Add one accordion open functionality

ivy-edp:
I think I can remove the summary from the accordion or should I leave it for now?


ivy-lli (approved):
Remove it 👍

PR #342, opened 2023-11-03
XIVY-12548 Add limit to allType-search

ivy-edp:
I have set the limit to 100 and reduced the waiting time to 150ms. With the limit, the search is much faster now. The part 'loading more types...' is now practically invisible to me.

### DIFF
--- a/packages/protocol/src/data/inscription.ts
+++ b/packages/protocol/src/data/inscription.ts
@@ -762,6 +762,7 @@ export interface ScriptingDataArgs {
 }
 export interface TypeSearchRequest {
   context: InscriptionContext;
+  limit: number;
   type: string;
 }
 export interface Void {}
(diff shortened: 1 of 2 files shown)


packages/protocol/src/data/inscription.ts
```diff
@@ -762,6 +762,7 @@ export interface ScriptingDataArgs {
 }
 export interface TypeSearchRequest {
   context: InscriptionContext;
+  limit: number;
```
ivy-lli, line 765: so the limit can no be defined by the client?

ivy-lli (approved):

packages/protocol/src/data/inscription.ts
ivy-edp: Yes, the client must specify a limit, otherwise no data will be returned.

PR #352, opened 2023-11-06
XIVY-12538 program 'userconfig' as map

ivy-rew:


ivy-lli (approved):

packages/editor/src/components/parts/program/configuration/ConfigurationPart.tsx
```diff
@@ -76,7 +76,7 @@ const ConfigurationPart = () => {
   return (
     <>
       {editorItems.map((widget, index) => (
-        <div key={index}>{renderWidgetComponent(widget)}</div>
+        <div key={index}>{renderWidgetComponent(widget, index.toString())}</div>
```
ivy-lli, line 79: not sure that this is correct. 
As far as I understood this, you should now have a id on the widget, which is corresponding to the key of the config data structure

packages/protocol/src/defaults/common-defaults.ts
```diff
@@ -231,6 +232,11 @@ export const DEFAULT_PROGRAM_INTERFACE_START_DATA: ProgramInterfaceStartData = {
   }
 } as const;
 
+export const DEFAULT_CONFIGURATION_DATA: ConfigurationData = {
+  javaClass: '',
+  userConfig: { '1': '' }
```
ivy-lli, line 237: i think the default of this is an empty object...

ivy-lli (changes requested):

packages/editor/src/components/parts/program/configuration/ConfigurationPart.test.tsx
```diff
@@ -31,7 +31,7 @@ describe('ConfigurationPart', () => {
 
   test('full data', async () => {
     renderPart({
-      userConfig: '123'
+      userConfig: { directory: '123' }
```
ivy-rew, line 34: directory : /tmp/myDir

packages/protocol/src/defaults/common-defaults.ts
```diff
@@ -234,7 +234,7 @@ export const DEFAULT_PROGRAM_INTERFACE_START_DATA: ProgramInterfaceStartData = {
 
 export const DEFAULT_CONFIGURATION_DATA: ConfigurationData = {
   javaClass: '',
-  userConfig: { '1': '' }
+  userConfig: { '': '' }
```
ivy-lli, line 237: why not {}?

ivy-lli (approved):

PR #362, opened 2023-11-09
Improve Design and Browserfunctionality

ivy-edp:
Added doubleclick-insert to cms- & datatype-browser. Should I also add it to the attribute-browser?
Improved table- and helpertextheight for the browsers
Adjusted the paddingleft for Treetables/removed blue cirle
Made simpleName in Datatype-Browser bold
Color-Changes: 
- Better contrast (as Sprint-Review suggested)
- New Hover and Selectstate-color "blue" for Input/ScriptInput/Combobox/checkbox/radio/table

### DIFF
--- a/packages/editor/src/components/browser/Browser.tsx
+++ b/packages/editor/src/components/browser/Browser.tsx
@@ -24,10 +24,19 @@ const Browser = ({ open, onOpenChange, types, accept, location, cmsOptions }: Br
   const { editorRef } = useEditorContext();
   const [active, setActive] = useState<BrowserType>(types[0]);
 
-  const attrBrowser = useAttributeBrowser(location);
-  const cmsBrowser = useCmsBrowser(cmsOptions);
+  const acceptBrowser = () => {
+    accept(allBrowsers.find(browser => browser.id === active)?.accept() ?? '', active);
+  };
+
+  const onRowDoubleClick = () => {
+    onOpenChange(false);
+    acceptBrowser();
+  };
+
+  const attrBrowser = useAttributeBrowser(onRowDoubleClick, location);
+  const cmsBrowser = useCmsBrowser(onRowDoubleClick, cmsOptions);
   const funcBrowser = useFuncBrowser();
-  const dataTypeBrowser = useDataTypeBrowser();
+  const dataTypeBrowser = useDataTypeBrowser(onRowDoubleClick);
   const catPathChooserBrowser = useCatPathChooserBrowser();
   const tableColBrowser = useTableColBrowser();
   const sqlOpBrowser = useSqlOpBrowser();
@@ -35,9 +44,6 @@ const Browser = ({ open, onOpenChange, types, accept, location, cmsOptions }: Br
   const allBrowsers = [attrBrowser, cmsBrowser, funcBrowser, dataTypeBrowser, catPathChooserBrowser, tableColBrowser, sqlOpBrowser];
 
   const tabs = allBrowsers.filter(browser => types.includes(browser.id));
-  const acceptBrowser = () => {
-    accept(allBrowsers.find(browser => browser.id === active)?.accept() ?? '', active);
-  };
 
   return (
     <>
--- a/packages/editor/src/components/widgets/table/cell/ExpandableCell.tsx
+++ b/packages/editor/src/components/widgets/table/cell/ExpandableCell.tsx
@@ -11,9 +11,18 @@ type ExpandableCellProps<TData> = {
   isUnknown?: boolean;
   icon?: IvyIcons;
   title?: string;
+  additionalInfo?: string;
 };
 
-export function ExpandableCell<TData>({ cell, isLoaded, loadChildren, isUnknown, icon, title }: ExpandableCellProps<TData>) {
+export function ExpandableCell<TData>({
+  cell,
+  isLoaded,
+  loadChildren,
+  isUnknown,
+  icon,
+  title,
+  additionalInfo
+}: ExpandableCellProps<TData>) {
   const row = cell.row;
   const onClick = () => {
     if (isLoaded === false && loadChildren) {
@@ -22,7 +31,7 @@ export function ExpandableCell<TData>({ cell, isLoaded, loadChildren, isUnknown,
     row.toggleExpanded(true);
   };
   return (
-    <div className='row-expand' style={{ paddingLeft: `${row.depth}rem` }} title={title}>
+    <div className='row-expand' style={{ paddingLeft: `calc(${row.depth}*(var(--tree-gap))` }} title={title}>
       {row.getCanExpand() ? (
         <>
           <Button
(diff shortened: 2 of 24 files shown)


packages/editor/src/components/widgets/table/cell/ExpandableCell.tsx
```diff
@@ -39,12 +40,18 @@ export function ExpandableCell<TData>({ cell, isLoaded, loadChildren, isUnknown,
...
+          <span className='row-expand-label'>
+            <b>{cell.getValue() as string}</b>
+          </span>
+          <span className='row-expand-label'>- {helperInfo}</span>
         </>
-      )}{' '}
-      <span className='row-expand-label'>{cell.getValue() as string}</span>
+      ) : (
+        <span className='row-expand-label'>{cell.getValue() as string}</span>
+      )}
```
ivy-lli, line 54: instead of this big conditional stuff move the additional info more to the place where it is needed, so you can reduce duplication!
And e.g. use css to make the difference between normal text and addional info

```jsx
<span className='row-expand-label'>{cell.getValue() as string}</span>
{helperInfo ?? <span className='row-expand-label-info'>- {helperInfo}</span>}
```

packages/editor/src/components/widgets/table/cell/ExpandableCell.tsx
```diff
@@ -11,9 +11,10 @@ type ExpandableCellProps<TData> = {
   isUnknown?: boolean;
   icon?: IvyIcons;
   title?: string;
+  helperInfo?: string;
 };
 
-export function ExpandableCell<TData>({ cell, isLoaded, loadChildren, isUnknown, icon, title }: ExpandableCellProps<TData>) {
+export function ExpandableCell<TData>({ cell, isLoaded, loadChildren, isUnknown, icon, title, helperInfo }: ExpandableCellProps<TData>) {
```
ivy-lli, line 17: don't really like the name helperinfo. additionalInfo or something like that is more suitable I think
And helperinfo is really near to the helper text in the browser

packages/editor/src/components/widgets/table/row/SelectRow.css
```diff
@@ -2,9 +2,16 @@
...
 
 .selectable-row:where([data-state='selected']) {
-  background-color: var(--N100);
+  background-color: var(--A400);
+  color: var(--background);
+}
+
+.selectable-row:where([data-state='selected']:last-child) {
+  background-color: var(--A400);
+  color: var(--background);
```
ivy-lli, line 15: duplicated from above -> not needed...

packages/editor/src/components/widgets/table/row/SelectRow.css
```diff
@@ -2,9 +2,16 @@
...
+  border-radius: var(--border-radius) !important;
```
ivy-lli, line 16: why is this needed???

packages/editor/src/components/widgets/table/row/SelectRow.tsx
```diff
@@ -6,13 +6,24 @@ type SelectRowProps<TData> = {
...
+      if (event.detail === 1) {
+        if (isNotSelectable) {
+          undefined;
+        } else {
+          row.getToggleSelectedHandler()(event);
+        }
+      } else if (onDoubleClick) {
+        onDoubleClick();
+      }
+    }}
```
ivy-lli: instead of undefined, turn the condition:
```ts
if (event.detail === 1) {
  if (!isNotSelectable) {
    row.getToggleSelectedHandler()(event);
  }
} else if (onDoubleClick) {
  onDoubleClick();
}
```

packages/editor/src/components/widgets/table/table/Table.css
```diff
@@ -19,3 +18,15 @@
 .table tr :where(td) {
   border-top: var(--seperator-border);
 }
+
+.table tr:last-child :where(td:first-child) {
```
ivy-lli, line 22: where is only needed if you use more than one selector in it
reduce complication -> `.table tr:last-child td:first-child {`

packages/editor/src/components/widgets/table/table/Table.css
```diff
@@ -19,3 +18,15 @@
...
+  border-bottom-left-radius: var(--border-radius);
+}
+
+.table tr:last-child :where(td:last-child) {
```
ivy-lli, line 26: same here

packages/editor/src/components/browser/Browser.tsx
```diff
@@ -17,27 +17,34 @@ type BrowserProps = UseBrowserReturnValue & {
...
+    onOpenChange(false);
+    acceptBrowser();
+  };
+
   const attrBrowser = useAttributeBrowser(location);
-  const cmsBrowser = useCmsBrowser(cmsOptions);
+  const cmsBrowser = useCmsBrowser(onBrowserRowDoubleClick, cmsOptions);
   const funcBrowser = useFuncBrowser();
-  const dataTypeBrowser = useDataTypeBrowser();
+  const dataTypeBrowser = useDataTypeBrowser(onBrowserRowDoubleClick);
```
ivy-lli, line 40: I know that you need to add this to the browsers, but honestly I think I would expect this behaviour on every browser!
Because otherwise, no one will understand why it will work on this browser but not on the other one...

ivy-lli (approved):
Looks good!
Some small comments below

And in general, I would only use optional attributes only if you have a use case. So first work with required params and if you have the use case, make it optional

packages/editor/src/components/widgets/table/row/SelectRow.css
ivy-edp: You are correct; it is no longer necessary. I encountered an issue where the border-radius from the table.css was not being applied. In an attempt to address this, I applied the following CSS. However, I have removed it, and everything is still works fine.

packages/editor/src/components/browser/Browser.tsx
ivy-edp: Yes, I agree with you. We can certainly brainstorm on how to solve this more effectively 👍

packages/editor/src/components/widgets/table/row/SelectRow.tsx
```diff
@@ -6,13 +6,24 @@ type SelectRowProps<TData> = {
...
+    onClick={event => {
+      if (event.detail === 1) {
+        if (!isNotSelectable) {
+          row.getToggleSelectedHandler()(event);
+        }
+      } else if (onDoubleClick) {
+        if (!isNotSelectable) {
+          onDoubleClick();
+        }
+      }
```
ivy-lli, line 25: move isnotselectable to the start, instead of check it in each if block

ivy-edp: 👍

PR #363, opened 2023-11-10
Enable eslint rule import/no-unresolved again

ivy-cst:
Is not a problem any more that eslint has troubles with the imports of esm bundles


ivy-lli (approved):
strange :D

PR #364, opened 2023-11-10
Don't compile test files

ivy-lli:
the path reference needs to direct to the build tsconfig. otherwise test files are compiled to the lib dir.


.gitignore
```diff
@@ -33,4 +33,4 @@ report.xml
 test-results
 playwright-report
 
-*tsconfig.tsbuildinfo
+*tsconfig.*tsbuildinfo
```
ivy-cst, line 36: `tsconfig.*tsbuildinfo` should be enough without `*`

ivy-cst (approved):
Cool :+1:

PR #369, opened 2023-11-10
Add more structure to tsconfig's

ivy-cst:


config/tsconfig.integration.json
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": "./tsconfig.base.json",
+  "compilerOptions": {
+    "moduleResolution": "Bundler",
```
ivy-cst, line 4: Set to bunlder according to: https://www.totaltypescript.com/tsconfig-cheat-sheet

integrations/standalone/tsconfig.json
```diff
@@ -1,20 +1,8 @@
...
-    "allowSyntheticDefaultImports": true,
-    "strict": true,
-    "forceConsistentCasingInFileNames": true,
-    "noFallthroughCasesInSwitch": true,
-    "module": "esnext",
-    "moduleResolution": "node",
-    "resolveJsonModule": true,
-    "isolatedModules": true,
-    "noEmit": true,
+    "noImplicitOverride": false,
```
ivy-cst, line 5: Had do change that for the integration, there is a lot of test code which does not declare the override explicitly. May be we should change that?

ivy-lli: why is the testcode a problem for the integration?
But yes I'm open for this change

ivy-lli (approved):
Looks good

integrations/standalone/tsconfig.json
ivy-cst: In my opinion in the end the noImplicitOverride should be removed here, and so true should be used for this flag as on the tsconfig.base. But at the moment a lot of test code does compile with this flag set to true.

ivy-lli: you mean playwright code? or jest code?

PR #389, opened 2023-11-16
XIVY-12800 Impl feedback for browsers

ivy-edp:
- **All Browsers**: Removed collapsible "Helper Text" in all Browsers. Now, the Helper Text is immediately displayed when an item is selected. If nothing is selected, the Helper is not shown.
image
- **CMS**: Added a refresh button in the top right corner to update the metadata if something new is added.
image
- **Datatype**: Removed static datatype. Now, at the beginning, nothing is displayed. When you search for something, it directly calls "allType." This simplifies the code, but I'm unsure if it is genuinely better for the user. Looking at it now, I would prefer that something like "most used" is displayed initially. What are your thoughts?
image


ivy-lli (approved):
I like the simplification of the browsers, especially of the datatype browser.
But personally, I don't like it that there is now nothing there...

But if we change that later "again" 🙈, we can simply return some types in the allTypes meta request, when the search is empty.

PR #391, opened 2023-11-17
XIVY-12800 Adjust color, radio/checkbox/datatypeb.

ivy-edp:
- Adjust the shades of blue
- Enhance the visibility of radio buttons and checkboxes in dark mode.
image
- Implemented feedback for the datatype browser: relocate the "Use Type as List" checkbox to the bottom and introduce an information cell indicating that entering a search term is required to display results.
image

### DIFF
--- a/packages/editor/src/components/browser/Browser.css
+++ b/packages/editor/src/components/browser/Browser.css
@@ -34,7 +34,12 @@
 }
 
 .insert {
-  color: var(--A400);
+  color: white;
+  background-color: var(--A300);
+}
+
+.insert:hover {
+  background-color: var(--A500);
 }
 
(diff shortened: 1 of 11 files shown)


packages/editor/src/components/browser/Browser.css
```diff
@@ -34,7 +34,12 @@
 }
 
 .insert {
-  color: var(--A400);
+  color: white;
```
ivy-lli, line 37: does this work in light and dark mode?

ivy-lli (approved):

packages/editor/src/components/browser/Browser.css
ivy-edp: Yes, i think so:
image
image

PR #392, opened 2023-11-17
XIVY-12684 Fix navigate to process in integration

ivy-lli:
This affects the link on the webservice start which changes the inscription view to the webservice process.

As the old solution changes the url of the window, this solution works in the standalone inscription view, but not integrated into the process editor.


ivy-cst:
Now it should work in all integrations, right?

ivy-cst (approved):

ivy-lli:
> Now it should work in all integrations, right?

I haven't test it yet, but as this now changes only the state of the current context, this should be perfectly working also in the process editor integration

ivy-edp (approved):
👍

PR #394, opened 2023-11-17
XIVY-12560 Only use pid where it is needed

ivy-lli:
Else it is in the tanstack query cache key, so if another element requests the same data, it can not be reused from the cache


ivy-cst (approved):
Nice :+1: 
There are some `const { elementContext: context }` left, can you replace all?

ivy-edp (approved):

ivy-lli:
> Nice 👍 There are some `const { elementContext: context }` left, can you replace all?

I replaced all which makes no difference, all others are inside of another object and are soone named 'context'. So I need to rename it somewhere, except we rename those in the backend.

PR #398, opened 2023-11-20
XIVY-12800 Change blue color

ivy-edp:


packages/editor/src/App.css
```diff
@@ -28,6 +29,9 @@
 
   --selected-text: #ffffff;
   --background-radio-checkbox: var(--A50);
+  --icon-fill-radio-checkbox: var(--A300);
+
+  --selectedRowFont: var(--background);
```
ivy-lli, line 34: please use constistent names -> --selected-row-font

And please name the variables like the content what they include, font is not the color for me, it is the font.
So this should be something like --selected-row-color (because color is the color of the font)

packages/editor/src/App.css
```diff
@@ -37,6 +41,7 @@
   --error-border: 1px solid var(--error-color);
   --seperator-border: var(--basic-border);
   --icon-seperator-line: 1px solid var(--N300);
+  --border-radio-checkbox: 1px solid var(--A75);
```
ivy-lli, line 44: I know not invented on this PR. But I recognized this now so can you improve this now or in the next PR?

Streamline the names of these variables, 'border' at the start or the end but not both.
--error-border
--border-radio-checkbox

ivy-lli: i would always put the css specific hint (what this variable is about) at the end.
E.g. radio-checkboy-border, radio-checkboy-background, warning-color, selected-text-color, etc.

ivy-lli (changes requested):

ivy-lli (approved):
😎

PR #415, opened 2023-11-24
Fix raise version script

ivy-cst:
- Because some pom.xml files where moved
- Before this error was just swallowed, now it is visible with 'set -e'


alexsuter (approved):

ivy-lli (commented):
Thanks

PR #417, opened 2023-11-24
Remove global button unset

ivy-lli:


packages/editor/src/components/editors/InscriptionEditor.css
```diff
@@ -1,15 +1,3 @@
-button {
-  all: unset;
-}
```
ivy-lli, line 3: bad in integration, because break other buttons outside of inscription view

ivy-edp (approved):

PR #422, opened 2023-11-24
Use package instead of build:production

ivy-cst:


ivy-cst:
Do you like package or not?

ivy-lli (approved):
ok for me. better than write build:production 🤣

PR #440, opened 2023-12-05
Stabilize e2e test

ivy-lli:
I think, I finally found one of the lines that causes so many tests to flakines...

The problem here was that the clean method of the table iterates over every row and removes them, by checking the count of the rows.
But the await on the remove call only waits until the "button is clicked", but this doesn't mean that the code behind this button removed it from the table. So probably the next count on the rows happens with the previous row still visible.
With the check that the row count needs to be the current count - 1 it waits until this condition is true or fails if it is not.


ivy-edp (approved):
nice 👍

weissreto (approved):
🙏

PR #447, opened 2023-12-06
XIVY-13052 Small monaco fixes

ivy-lli:
- Fix monaco selection in popup
- Init height of script code block equals code
- Browser button in script-area should always be over scrollbar


ivy-edp (approved):
👍

PR #448, opened 2023-12-06
XIVY-12992 Improve Browsers & adjust Typebrowser

ivy-edp:
Adjusted Typebrowser:
- By default, only the most frequently used types (ivytypes and a few others) and data classes are displayed (sorted alphabetically, first by full qualified name and then by simple name).
- There is now a checkbox "Search over all types." When this is activated, the search is conducted over all types and the ivytypes/dataclasses are no longer displayed; otherwise, the search is limited to data classes and ivytypes.
- Open ToDos: Reguel still needs to create metadata for the type browser so that in addition to data classes and ivytypes, own custom-created classes/types are also visible. It will also be nessesary that two icons are designed, to indicate whether the type is a data class, an Ivy type, or a "normal type." Possibly, an icon for the own custom classes might also make sense.
grafik

- Icons have been added to the known browser tabs.
- Icons have also been added to individual rows in the attribute/cms and type browser. The padding in the tree table of the CMS browser has also been improved.
- Initially, focus is always set on the search field.
grafik

### DIFF
--- a/packages/editor/src/components/browser/CmsBrowser.tsx
+++ b/packages/editor/src/components/browser/CmsBrowser.tsx
@@ -31,7 +31,8 @@ export const useCmsBrowser = (onDoubleClick: () => void, options?: CmsOptions): 
         onDoubleClick={onDoubleClick}
       />
     ),
-    accept: () => value
+    accept: () => value,
+    icon: IvyIcons.Cms
   };
 };
 
@@ -64,7 +65,13 @@ const CmsBrowser = ({ value, onChange, noApiCall, typeFilter, onDoubleClick }: C
             <ExpandableCell
               cell={cell}
               title={cell.row.original.name}
-              icon={cell.row.original.type === 'FOLDER' ? IvyIcons.GoToSource : undefined}
+              icon={
+                cell.row.original.type === 'FOLDER'
+                  ? IvyIcons.FolderOpen
+                  : cell.row.original.type === 'FILE'
+                  ? IvyIcons.File
+                  : IvyIcons.ChangeType
+              }
               additionalInfo={cell.row.original.type}
             />
           );
--- a/packages/editor/src/components/browser/TypeBrowser.tsx
+++ b/packages/editor/src/components/browser/TypeBrowser.tsx
@@ -5,6 +5,7 @@ import type { ColumnDef, ExpandedState, FilterFn, RowSelectionState } from '@tan
 import { flexRender, getCoreRowModel, getExpandedRowModel, getFilteredRowModel, useReactTable } from '@tanstack/react-table';
 import { useEditorContext, useMeta } from '../../context';
 import type { JavaType } from '@axonivy/inscription-protocol';
+import { IvyIcons } from '@axonivy/editor-icons';
 export const TYPE_BROWSER_ID = 'type' as const;
 
 export const useTypeBrowser = (onDoubleClick: () => void): UseBrowserImplReturnValue => {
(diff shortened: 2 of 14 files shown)


packages/editor/src/components/browser/CmsBrowser.tsx
```diff
@@ -64,7 +65,7 @@ const CmsBrowser = ({ value, onChange, noApiCall, typeFilter, onDoubleClick }: C
             <ExpandableCell
               cell={cell}
               title={cell.row.original.name}
-              icon={cell.row.original.type === 'FOLDER' ? IvyIcons.GoToSource : undefined}
+              icon={cell.row.original.type === 'FOLDER' ? IvyIcons.FolderOpen : IvyIcons.File}
```
ivy-lli, line 68: what about content object strings?

packages/editor/src/components/browser/TypeBrowser.tsx
```diff
@@ -13,40 +14,99 @@ export const useTypeBrowser = (onDoubleClick: () => void): UseBrowserImplReturnV
...
+type TypeKind = 'DATACLASS' | 'NORMALTYPE' | 'OWNTYPE';
+
 const TypeBrowser = (props: { value: string; onChange: (value: string) => void; onDoubleClick: () => void }) => {
+  type TypeBrowserObject = JavaType & {
+    kind: TypeKind;
+  };
+
   const { context } = useEditorContext();
 
+  const [allSearchActive, setAllSearchActive] = useState<boolean>(false);
```
ivy-lli, line 31: <boolean> is not needed if you set the default value to false -> `useState(false)`

packages/editor/src/components/browser/useBrowser.ts
```diff
@@ -20,6 +21,7 @@ export type BrowserType =
 export type UseBrowserImplReturnValue = Omit<Tab, 'id'> & {
   id: BrowserType;
   accept: () => string;
+  icon?: IvyIcons;
```
ivy-lli, line 24: if all browser now support an icon, you can also make this required -> remove optional

packages/editor/src/components/widgets/input/IconInput.tsx
```diff
@@ -3,17 +3,30 @@ import type { IvyIcons } from '@axonivy/editor-icons';
...
+  const inputRef = useRef<HTMLInputElement>(null);
+
+  useEffect(() => {
+    if (initFocus) {
+      const timer = setTimeout(() => {
+        inputRef.current?.focus();
+      }, 0);
+      return () => clearTimeout(timer);
+    }
+    return () => {};
```
ivy-lli, line 23: I think you don't need a return value if nothing is returned. -> remove return if initFocus is not true

packages/editor/src/components/widgets/input/IconInput.tsx
```diff
@@ -3,17 +3,30 @@ import type { IvyIcons } from '@axonivy/editor-icons';
...
-import { forwardRef } from 'react';
+import { forwardRef, useEffect, useRef } from 'react';
 
 export type IconInputProps = InputProps & {
   icon: IvyIcons;
+  initFocus?: boolean;
 };
 
-const IconInput = forwardRef<HTMLInputElement, IconInputProps>(({ icon, ...props }, forwardedRef) => {
+const IconInput = forwardRef<HTMLInputElement, IconInputProps>(({ icon, initFocus, ...props }, forwardedRef) => {
```
ivy-lli, line 13: is this iconInput used somewhere with a ref? because if not you can also remove this forwardRef stuff.

packages/editor/src/components/widgets/tab/Tab.css
```diff
@@ -50,3 +53,8 @@
   flex-direction: column;
   gap: var(--size-3);
 }
+
+i {
+  padding-top: 1px;
+  padding-right: var(--size-1);
+}
```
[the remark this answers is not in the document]
ivy-lli: And anyway are you sure that this is the best solution for this tab icon?
I think just add a gap to the tab-trigger is better than a padding-right. And the padding-top is maybe not even needed right? (maybe just from the zoom)

packages/editor/src/components/browser/TypeBrowser.tsx
ivy-lli, line 1: I need to look at this more tomorrow...

ivy-lli (changes requested):

packages/editor/src/components/browser/CmsBrowser.tsx
ivy-edp: I believe Anais has not yet designed a suitable icon for it, or is the following icon intended for strings?
grafik

packages/editor/src/components/browser/TypeBrowser.tsx
```diff
@@ -150,7 +220,9 @@ const TypeBrowser = (props: { value: string; onChange: (value: string) => void;
           </tbody>
         ) : (
           <tbody>
-            <TableCell>Please enter a search term to see results</TableCell>
+            <tr>
+              <TableCell>Please enter a search term to see results</TableCell>
+            </tr>
```
ivy-lli, line 225: Nice! Also saw this error in the console log in the past, but forgot again :)

packages/editor/src/components/browser/useBrowser.ts
ivy-edp: Thank you, I also considered that. However, I left it optional because, as far as I could see, Anais has not designed an icon for, for example, the Category Path Chooser or the Table Column Browser yet. I will make it required, though, and temporarily use an icon that seems reasonably fitting to me for these browsers.

packages/editor/src/components/browser/TypeBrowser.tsx
```diff
@@ -137,7 +207,7 @@ const TypeBrowser = (props: { value: string; onChange: (value: string) => void;
           }
         }}
       >
-        {mainFilter.length > 0 ? (
+        {(mainFilter.length > 0 && allSearchActive) || !allSearchActive ? (
```
ivy-lli, line 210: instead of this duplication of logic, what should be shown if the allsearch is active or not, can you just show this empty message if the rowmodel is empty? Or does this not work? 
And of course then the message would be a little bit different, more like 'No type found, use the search for ...'

packages/editor/src/components/browser/TypeBrowser.tsx
```diff
@@ -13,40 +14,99 @@ export const useTypeBrowser = (onDoubleClick: () => void): UseBrowserImplReturnV
...
+        return a.simpleName.localeCompare(b.simpleName);
+      });
+
+      const sortedMappedDataClasses = mappedDataClasses.slice().sort((a, b) => {
+        const fqCompare = a.fullQualifiedName.localeCompare(b.fullQualifiedName);
+        if (fqCompare !== 0) {
+          return fqCompare;
+        }
+        return a.simpleName.localeCompare(b.simpleName);
+      });
```
ivy-lli, line 79: deduplicate this compare logic into a shared function.
A cool improvement, but I'm not sure if the backend shouldn't already deliver this types sorted. But ok for now to leave this code here in the client :)

packages/editor/src/components/browser/TypeBrowser.tsx
```diff
@@ -13,40 +14,99 @@ export const useTypeBrowser = (onDoubleClick: () => void): UseBrowserImplReturnV
     id: TYPE_BROWSER_ID,
     name: 'Type',
     content: <TypeBrowser value={value} onChange={setValue} onDoubleClick={onDoubleClick} />,
-    accept: () => value
+    accept: () => value,
+    icon: IvyIcons.DataClass
   };
 };
 
+type TypeKind = 'DATACLASS' | 'NORMALTYPE' | 'OWNTYPE';
```
ivy-lli, line 22: normally I would say this kind is perfect. But as far as I understand you code, you currently only need this to differ the icons right?
If this is the case I think this is a little "overkill" and more or less useless code which needs to run. Better just add directly the correct icon to the TypeBrowserObject.

packages/editor/src/components/browser/TypeBrowser.tsx
```diff
@@ -13,40 +14,88 @@ export const useTypeBrowser = (onDoubleClick: () => void): UseBrowserImplReturnV
...
   useEffect(() => {
-    if (mainFilter.length > 0) {
-      allDatatypes.sort((a, b) => a.simpleName.localeCompare(b.simpleName));
+    const sortAndMapTypes = (types: TypeBrowserObject[]): TypeBrowserObject[] => {
+      return types.slice().sort((a, b) => {
+        const fqCompare = a.fullQualifiedName.localeCompare(b.fullQualifiedName);
+        if (fqCompare !== 0) {
+          return fqCompare;
+        }
+        return a.simpleName.localeCompare(b.simpleName);
```
ivy-lli, line 49: of course you can also create on function that do the sorting for you. I expected only the sorting function itself as shared function, but ok

```ts
const typeComparator = (a: TypeBrowserObject, b: TypeBrowserObject) => {
  const fqCompare = a.fullQualifiedName.localeCompare(b.fullQualifiedName);
  if (fqCompare !== 0) {
    return fqCompare;
  }
  return a.simpleName.localeCompare(b.simpleName);
}

const sortedIvyTypes = mappedIvyTypes.sort(typeComparator);
const sortedMappedDataClasses = mappedDataClasses.sort(typeComparator);
```

packages/editor/src/components/browser/TypeBrowser.tsx
```diff
@@ -137,7 +196,7 @@ const TypeBrowser = (props: { value: string; onChange: (value: string) => void;
           }
         }}
       >
-        {mainFilter.length > 0 ? (
+        {tableDynamic.getRowModel().rows.length > 0 ? (
           <tbody>
```
ivy-lli, line 200: you can also deduplicate the tbody by move this out of the condition. maybe then you need to use a <> fragment inside

ivy-lli (approved):
Looks good, thanks for the improvements.
Two small comments which you may want to change

PR #454, opened 2023-12-11
XIVY-13055 Add selected Text to search in Typebrow

ivy-edp:
I hope my approach is OK. there is already the functionality that the selected text is replaced but I have not seen anywhere that the content of the selected text is already saved somewhere.

### DIFF
--- a/integrations/standalone/tests/mock/browser.spec.ts
+++ b/integrations/standalone/tests/mock/browser.spec.ts
@@ -34,6 +34,24 @@ test.describe('Script browser', () => {
     await expect(code(page).getByRole('textbox')).toHaveValue('test 123 <%=in.bla%>');
   });
 
+  test('browser init searchfilter', async ({ page }) => {
+    const inscriptionView = await InscriptionView.mock(page);
+    const task = inscriptionView.accordion('Call');
+    await task.toggle();
+
+    await page.getByText('Code').click();
+    const codeField = task.scriptArea();
+    await codeField.fill('handle');
+
+    await code(page).dblclick();
+    await browserBtn(page).click();
+    await expect(page.getByRole('dialog')).toBeVisible();
+
+    await page.getByText('Type').last().click();
+
+    await expect(page.getByPlaceholder('Search')).toHaveValue('handle');
+  });
+
   test('browser add cms string', async ({ page }) => {
     const inscriptionView = await InscriptionView.mock(page);
     const task = inscriptionView.accordion('Task');
--- a/packages/editor/src/components/browser/Browser.tsx
+++ b/packages/editor/src/components/browser/Browser.tsx
@@ -19,9 +19,10 @@ type BrowserProps = UseBrowserReturnValue & {
   accept: (value: string, type: BrowserType) => void;
   location: string;
   cmsOptions?: CmsOptions;
+  initSearchFilter?: () => string;
 };
 
-const Browser = ({ open, onOpenChange, types, accept, location, cmsOptions }: BrowserProps) => {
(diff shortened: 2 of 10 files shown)


packages/editor/src/components/browser/Browser.tsx
```diff
@@ -46,8 +48,13 @@ const Browser = ({ open, onOpenChange, types, accept, location, cmsOptions }: Br
 
   const tabs = allBrowsers.filter(browser => types.includes(browser.id));
 
+  const handleTextSelection = () => {
+    const newSelectedText = window.getSelection()?.toString();
+    setSelectedText(newSelectedText === undefined ? '' : newSelectedText);
+  };
+
```
ivy-lli, line 55: I think it should be possible to reuse the logic for the replacement for this selected text...

ivy-lli (changes requested):

packages/editor/src/components/browser/Browser.tsx
ivy-edp: I have now adapted the entire thing so that a new function called "getMonacoSelection" is available in useModifyEditor, which returns the value of the current selection. I then pass this method to the browser, and it calls it when the browser icon is clicked. I have added the functionality only for ScriptArea fields for now.

ivy-lli (changes requested):
Nice work. Some improvement suggestions.
And I think a mock test should be added here...

packages/editor/src/components/browser/Browser.tsx
```diff
@@ -50,7 +53,11 @@ const Browser = ({ open, onOpenChange, types, accept, location, cmsOptions }: Br
     <>
       <Dialog open={open} onOpenChange={onOpenChange}>
         <DialogTrigger asChild>
-          <Button icon={IvyIcons.ListSearch} aria-label='Browser' />
+          <Button
+            icon={IvyIcons.ListSearch}
+            aria-label='Browser'
+            onClick={() => initSearchFilter !== undefined && setInitSearchFilterValue(initSearchFilter)}
```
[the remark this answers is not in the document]
ivy-edp: I was initially also a bit skeptical about whether this would work, but it with fortunatly the onClick worked well.

I tried it this way, but unfortunately, it didn't work. I believe the issue is that I lose the selection after clicking the button to open the browser. Therefore, when I pass the initSearchFilter function to useTypeBrowser and call it during the initialization of the browser, it returns an empty string because the selection has already been cleared by the button click.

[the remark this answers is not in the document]
ivy-edp: Nevermind,  it worked as you mentioned. I made a mistake during testing and tried it on a ScriptInput instead of a ScriptArea :-)

integrations/standalone/tests/mock/browser.spec.ts
```diff
@@ -34,6 +34,32 @@ test.describe('Script browser', () => {
...
+
+    await codeField.focus();
+    await assertCodeVisible(page);
+    await page.keyboard.down('Control');
+    await page.keyboard.press('A');
+    await page.keyboard.up('Control');
+
+    await page.keyboard.press('Delete');
+
+    await page.keyboard.type('handle');
```
ivy-lli, line 53: Maybe 'control' will not work in webkit, because webkit simulates mac a bit and on mac it is 'cmd+A', btw 'Meta'.

But I think the codeField (ScriptArea pageobject) has a clear of a fill method to clean/fill content... Why you don't reuse them?

ivy-lli (approved):

PR #455, opened 2023-12-13
XIVY-12992 Add new Icons and fix browser animation

ivy-edp:
I have added two new icons to the type browser. Now they are assigned as follows:
-     ownTypes and "normal" types are represented with the C-icon.
-     DataClasses are represented with the D-icon.
-     All types containing 'ivy' in the FullQualifiedName are represented with the ivy-icon.
What do you think about the icons?
grafik

Additionally, I ensured in the useEffect that there are no duplicated types in the typebrowser. Also I had to increase the timeout for focus in the browser for the search, as otherwise, the ease-in animation does not work.


packages/editor/src/components/browser/TypeBrowser.tsx
```diff
@@ -55,22 +55,31 @@ const TypeBrowser = ({ value, onChange, onDoubleClick, initSearchFilter }: TypeB
...
 
     if (allSearchActive) {
+      let uniqueAllTypes: JavaType[] = [];
+
       if (mainFilter.length > 0) {
-        allDatatypes.sort((a, b) => a.simpleName.localeCompare(b.simpleName));
+        uniqueAllTypes = Array.from(new Set(allDatatypes.map(datatype => datatype.fullQualifiedName)))
+          .map(fullQualifiedName => allDatatypes.find(datatype => datatype.fullQualifiedName === fullQualifiedName))
+          .filter(type => type !== undefined) as JavaType[];
+        uniqueAllTypes.sort((a, b) => a.simpleName.localeCompare(b.simpleName));
```
ivy-lli, line 64: solve this in backend

ivy-lli (changes requested):

ivy-lli (approved):

PR #456, opened 2023-12-13
Improve type cms browsers

ivy-edp:
I would now merge the corresponding changes to the release/11.2 branch. is this correct? In order for the type browser in particular to work, a new meta must also be merged down to 11.2.1 in the core. This would be done by reguel, but probably not today. can I merge the branch anyway or do i have to wait on the core-merge?

Merge the following features to 11.2.1:
- Browser tabs with icons
- Focus on search for browsers
- Icons in the attribute and CMS treetable and Type-List
- Improved the treetable indentation
- Type browser: The following types are now suggested at the start: ownTypes, DataClasses, IvyType, oftenUsed
- Typebrowser: You now have to consciously activate allType-Search to search for everything. Otherwise only the suggested types are searched for
- Typebrowser: Add selected Text to search in Typebrowser
- Bugfix: CMS behavior in mail same as with old Inscription View


ivy-lli (approved):
Does it work without the meta in the core -> merge it
Does it not work (crash UI or something like that) -> wait for core

PR #458, opened 2023-12-18
XIVY-12992 Fix defaulttypes for typebrowser

ivy-edp:
Modified the default types in the Typebrowser as follows:
- Within code blocks, the default types are prioritized in the following order: 1. Own Types, 2. Data Classes, 3. Ivy Types (with the condition that, if applied, there should be no "import..." statement and also sorted alphabetically but all elements from java.lang/util should be displayed first).
- In all other sections, particularly in the mapping table, the default types are ordered as follows: 1. Data Classes, 2. Ivy Types.

### DIFF
--- a/packages/editor/src/components/browser/TypeBrowser.tsx
+++ b/packages/editor/src/components/browser/TypeBrowser.tsx
@@ -6,16 +6,26 @@ import { flexRender, getCoreRowModel, getExpandedRowModel, getFilteredRowModel, 
 import { useEditorContext, useMeta } from '../../context';
 import type { JavaType } from '@axonivy/inscription-protocol';
 import { IvyIcons } from '@axonivy/editor-icons';
+import type { BrowserValue } from './Browser';
+import { getCursorValue } from './typeBrowser-utils';
 export const TYPE_BROWSER_ID = 'type' as const;
 
-type TypeBrowserObject = JavaType & { icon: IvyIcons };
+export type TypeBrowserObject = JavaType & { icon: IvyIcons };
 
-export const useTypeBrowser = (onDoubleClick: () => void, initSearchFilter: () => string): UseBrowserImplReturnValue => {
-  const [value, setValue] = useState('type');
+export const useTypeBrowser = (onDoubleClick: () => void, initSearchFilter: () => string, location: string): UseBrowserImplReturnValue => {
+  const [value, setValue] = useState<BrowserValue>({ cursorValue: '' });
   return {
     id: TYPE_BROWSER_ID,
     name: 'Type',
-    content: <TypeBrowser value={value} onChange={setValue} onDoubleClick={onDoubleClick} initSearchFilter={initSearchFilter} />,
+    content: (
+      <TypeBrowser
+        value={value.cursorValue}
+        onChange={setValue}
+        onDoubleClick={onDoubleClick}
+        location={location}
+        initSearchFilter={initSearchFilter}
+      />
+    ),
     accept: () => value,
     icon: IvyIcons.DataClass
   };
@@ -23,12 +33,13 @@ export const useTypeBrowser = (onDoubleClick: () => void, initSearchFilter: () =
 
 interface TypeBrowserProps {
   value: string;
-  onChange: (value: string) => void;
+  onChange: (value: BrowserValue) => void;
   onDoubleClick: () => void;
   initSearchFilter: () => string;
+  location: string;
 }
 
-const TypeBrowser = ({ value, onChange, onDoubleClick, initSearchFilter }: TypeBrowserProps) => {
+const TypeBrowser = ({ value, onChange, onDoubleClick, initSearchFilter, location }: TypeBrowserProps) => {
   const { context } = useEditorContext();
 
   const [allSearchActive, setAllSearchActive] = useState(false);
@@ -52,6 +63,18 @@ const TypeBrowser = ({ value, onChange, onDoubleClick, initSearchFilter }: TypeB
         return fqCompare;
       }
       return a.simpleName.localeCompare(b.simpleName);
+    };
+
+    const ivyTypeComparator = (a: TypeBrowserObject, b: TypeBrowserObject) => {
+      const aHasJava = a.fullQualifiedName.startsWith('java.lang');
+      const bHasJava = b.fullQualifiedName.startsWith('java.lang');
+      if (aHasJava && !bHasJava) {
+        return -1;
+      }
+      if (!aHasJava && bHasJava) {
+        return 1;
+      }
+      return typeComparator(a, b);
     };
 
     if (allSearchActive) {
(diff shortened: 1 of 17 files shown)


packages/editor/src/components/widgets/code-editor/useCodeEditor.ts
```diff
@@ -14,6 +15,7 @@ export type ModifyAction = (value: string) => string;
 
 export const useMonacoEditor = (options?: { modifyAction?: ModifyAction; scriptAreaDatatypeEditor?: boolean }) => {
   const [editor, setEditor] = useState<monaco.editor.IStandaloneCodeEditor>();
+  const ivyTypes = useMeta('meta/scripting/ivyTypes', undefined, []).data;
```
ivy-lli, line 18: I don't like this!
Now everywhere this hook is used, this request is fired to the backend, even if you never use it...
If we make such conditions, make them based on the location or add more information besides the type to the modifyEditor call.

packages/editor/src/components/browser/TypeBrowser.tsx
```diff
@@ -80,19 +98,23 @@ const TypeBrowser = ({ value, onChange, onDoubleClick, initSearchFilter }: TypeB
...
+        const mappedOwnTypes: TypeBrowserObject[] = ownTypesWithoutDataClasses.map<TypeBrowserObject>(ownType => ({
+          icon: IvyIcons.DataClass,
+          ...ownType
+        }));
+        const sortedMappedOwnTypes = mappedOwnTypes.sort(typeComparator);
 
-      const sortedTypes: TypeBrowserObject[] = sortedMappedOwnTypes.concat(sortedMappedDataClasses).concat(sortedIvyTypes);
-      setTypes(sortedTypes);
+        const sortedTypes: TypeBrowserObject[] = sortedMappedOwnTypes.concat(sortedMappedDataClasses).concat(sortedIvyTypes);
+        setTypes(sortedTypes);
```
ivy-lli, line 111: why save it first in a variable?
`setTypes(sortedMappedOwnTypes.concat(sortedMappedDataClasses).concat(sortedIvyTypes));`

packages/editor/src/components/browser/TypeBrowser.tsx
```diff
@@ -46,7 +55,16 @@ const TypeBrowser = ({ value, onChange, onDoubleClick, initSearchFilter }: TypeB
...
+    const typeComparator = (a: TypeBrowserObject, b: TypeBrowserObject, ivyType?: boolean) => {
+      if (ivyType) {
+        const aHasJava = a.fullQualifiedName.includes('java.');
+        const bHasJava = b.fullQualifiedName.includes('java.');
+        if (aHasJava && !bHasJava) {
+          return -1;
+        } else if (!aHasJava && bHasJava) {
+          return 1;
+        }
+      }
```
ivy-lli, line 67: Never do this! Deduplicate your code is good, but with this additional parameter you change the whole function, so create a new function for this case, because it has nothing do to with the other one.

If you write an if with a return state, you can also write the else if as if, especially if you don't write an else case.

```ts
const ivyTypeComparator = (a: TypeBrowserObject, b: TypeBrowserObject) => {
  const aHasJava = a.fullQualifiedName.includes('java.');
  const bHasJava = b.fullQualifiedName.includes('java.');
  if (aHasJava && !bHasJava) {
    return -1;
  }
  if (!aHasJava && bHasJava) {
    return 1;
  }
  return typeComparator(a, b);
}
```

ivy-lli: And are you shure that 'java.' is the best to discover ivy types?
I can imagine that someone else also writes a package name with 'java.' in it or that our types have 'ivy' in it or something like that

ivy-lli (changes requested):

packages/editor/src/components/browser/TypeBrowser.tsx
ivy-edp: Thanks, from my point of view this should not be a problem as this (ivy)typeComparator is currently only used for the ivyTypes array and this remains static.

packages/editor/src/components/widgets/code-editor/useCodeEditor.ts
```diff
@@ -15,29 +15,34 @@ export type ModifyAction = (value: string) => string;
...
-    if (type === 'type' && options?.scriptAreaDatatypeEditor) {
-      const textWithLineBreak = 'import ' + value + ';\n';
-      const parts = value.split('.');
-      const shortValue = parts.length > 1 ? parts.pop() : value;
+    if (value.firstLineValue) {
+      let cursorValue = value.cursorValue;
+      if (options?.scriptAreaDatatypeEditor) {
+        const parts = cursorValue.split('.');
+        cursorValue = parts.length > 1 ? String(parts.pop()) : cursorValue;
+      }
```
ivy-edp, line 31: Theoretically, I could also perform this check in the TypeBrowser and completely remove the flag scriptAreaDatatypeEditor. However, in the TypeBrowser, I would do this using location.includes('code'). You can find this check already in the TypeBrowser, as I used it there to ensure that "import" is only created in the code block. However, I'm not sure if using .includes('code') covers all cases or allows for too many cases. @ivy-lli What do you think?
Also I removed the BrowserType because I currently don't need it for any condition.

packages/editor/src/components/browser/TypeBrowser.tsx
```diff
@@ -160,16 +184,32 @@ const TypeBrowser = ({ value, onChange, onDoubleClick, initSearchFilter }: TypeB
...
+        onChange({
+          cursorValue: cursorValueShort,
+          firstLineValue: 'import ' + helperText + ';\n'
+        });
+      } else {
+        onChange({ cursorValue: cursorValueShort });
+      }
+    } else {
+      onChange({ cursorValue: helperText });
+    }
```
ivy-edp, line 211: @ivy-lli What do you think about this part of the code? I find it a bit complicated, but I don't know, how to solve it differently. With this code, the import works much better than before.

ivy-lli: Don't understand the code here...
why do you need to extract the values from the helpertext and no longer from the full row with all the different properties?
And in general, I think this code here should be in a separate function and should have tests... Would also be much easier to read this logic and to check that it always does what we are expecting.

packages/editor/src/components/browser/TypeBrowser.tsx
```diff
@@ -175,12 +174,28 @@ const TypeBrowser = ({ onChange, onDoubleClick, initSearchFilter, location }: Ty
...
+          prefix = typeAsList ? 'List<' : '';
+          suffix = typeAsList ? value.fullQualifiedName + '>' : value.fullQualifiedName;
+        } else {
+          prefix = typeAsList ? 'java.util.List<' : '';
+          suffix = typeAsList ? value.fullQualifiedName + '>' : value.fullQualifiedName;
+        }
       }
+
+      return `${prefix}${suffix}`;
     };
```
ivy-lli, line 199: If you name this function 'addListGeneric' it should do exactly this, means, the typeAsList should not be a parameter, but a condition if this method is called or not. Then you can write this much simpler and much more readable:

```ts
const addListGeneric = (value: TypeBrowserObject, typeAsList: boolean, isIvyType: boolean, inCodeBlock: boolean) => {
  if (isIvyType && inCodeBlock) {
    return `List<${value.simpleName}>`;
  } 
  if (isIvyType && !inCodeBlock) {
    return `java.util.List<${value.simpleName}>`;
  }
  if (!isIvyType && inCodeBlock) {
    return `List<${value.fullQualifiedName}>`;
  }
  return `java.util.List<${value.fullQualifiedName}>`; 
}
```

or 

```ts
const addListGeneric = (value: TypeBrowserObject, typeAsList: boolean, isIvyType: boolean, inCodeBlock: boolean) => {
  let listPrefix = 'java.util.List';
  if (inCodeBlock) {
    listPrefix = 'List';
  }
  if (isIvyType) {
    return `${listPrefix}<${value.simpleName}>`;
  } 
  return `${listPrefix}<${value.fullQualifiedName}>`; 
}
```

packages/editor/src/components/browser/TypeBrowser.test.tsx
ivy-lli, line 1: Good, but I thought more of move the function out of the react component maybe into a own hook if needed and test only this. Without meta without table, test only this function e.g. what if you call the function which isInCodeblock with an ivy type.

ivy-lli (changes requested):

ivy-lli (approved):
Nice 💯

PR #459, opened 2023-12-19
XIVY-12549 Add simple functionbrowser

ivy-edp:
grafik

### DIFF
--- a/packages/editor/src/components/browser/FunctionBrowser.tsx
+++ b/packages/editor/src/components/browser/FunctionBrowser.tsx
@@ -1,20 +1,158 @@
-import { useState } from 'react';
-import { Input } from '../widgets';
+import { useEffect, useMemo, useState } from 'react';
+import { ExpandableCell, SelectRow, Table, TableCell } from '../widgets';
 import type { UseBrowserImplReturnValue } from './useBrowser';
+import type { Function } from '@axonivy/inscription-protocol';
+import {
+  useReactTable,
+  type ColumnDef,
+  type ExpandedState,
+  type RowSelectionState,
+  getCoreRowModel,
+  getExpandedRowModel,
+  getFilteredRowModel,
+  flexRender,
+  type Row
+} from '@tanstack/react-table';
 import { IvyIcons } from '@axonivy/editor-icons';
+import { useEditorContext, useMeta } from '../../context';
 export const FUNCTION_BROWSER_ID = 'func' as const;
 
-export const useFuncBrowser = (): UseBrowserImplReturnValue => {
+export const useFuncBrowser = (onDoubleClick: () => void): UseBrowserImplReturnValue => {
   const [value, setValue] = useState('function');
   return {
     id: FUNCTION_BROWSER_ID,
     name: 'Function',
-    content: <FunctionBrowser value={value} onChange={setValue} />,
+    content: <FunctionBrowser value={value} onChange={setValue} onDoubleClick={onDoubleClick} />,
     accept: () => value,
     icon: IvyIcons.Function
   };
 };
 
-const FunctionBrowser = (props: { value: string; onChange: (value: string) => void }) => {
-  return <Input {...props} />;
+const FunctionBrowser = (props: { value: string; onChange: (value: string) => void; onDoubleClick: () => void }) => {
+  const { context } = useEditorContext();
+  const [method, setMethod] = useState('');
+  const [paramTypes, setParamTypes] = useState<string[]>([]);
+  const [type, setType] = useState('');
+
+  const { data: tree } = useMeta('meta/scripting/functions', undefined, []);
+  const { data: doc } = useMeta('meta/scripting/apiDoc', { context, method, paramTypes, type }, '');
+
+  const [selectedFunctionDoc, setSelectedFunctionDoc] = useState('');
+
+  const [showHelper, setShowHelper] = useState<boolean>(false);
+
+  const columns = useMemo<ColumnDef<Function>[]>(
+    () => [
+      {
+        accessorFn: row =>
+          `${row.name.split('.').pop()}${
+            row.isField === false ? `(${row.params.map(param => param.type.split('.').pop()).join(', ')})` : ''
+          }`,
+        id: 'name',
+        cell: cell => {
+          return (
+            <ExpandableCell
+              cell={cell}
+              title={cell.row.original.name}
+              icon={cell.row.original.isField ? IvyIcons.FolderOpen : IvyIcons.Function}
+              additionalInfo={cell.row.original.returnType ? cell.row.original.returnType.simpleName : 'no return type defined'}
+            />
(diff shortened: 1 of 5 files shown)


ivy-edp:
Currently we have solved the whole thing in such a way that I receive the complete tree (i guess approx. 13000 entries) statically from the backend. Basically, loading the functionbrowser is fast, but the search is currently still very slow, because to find an element in the entire tree, I first have to expand it completely.

packages/editor/src/components/widgets/code-editor/useCodeEditor.ts
```diff
@@ -39,6 +39,38 @@ export const useMonacoEditor = (options?: { modifyAction?: ModifyAction; scriptA
...
       const text = value.length > 0 && options?.modifyAction ? options.modifyAction(value) : value;
       editor.executeEdits('browser', [{ range: selection, text, forceMoveMarkers: true }]);
+      if (type === 'func') {
+        let lengthUntilOpeningParenthesis = 0;
+        for (let i = 0; i < text.length; i++) {
+          if (text[i] === '(') {
+            break;
+          }
+          lengthUntilOpeningParenthesis++;
+        }
```
ivy-lli, line 49: why not use text.indexOf('(')

packages/editor/src/components/widgets/code-editor/useCodeEditor.ts
```diff
@@ -39,6 +39,38 @@ export const useMonacoEditor = (options?: { modifyAction?: ModifyAction; scriptA
...
+            startLineNumber: startLineNumber,
+            startColumn: startColumn,
+            endLineNumber: endLineNumber,
+            endColumn: endColumn
+          };
+
+          setTimeout(() => {
+            editor.setSelection(textrange);
+            editor.focus();
+          }, 500);
```
ivy-lli, line 71: why is this timeout needed?

packages/editor/src/components/widgets/code-editor/useCodeEditor.ts
```diff
@@ -39,6 +39,38 @@ export const useMonacoEditor = (options?: { modifyAction?: ModifyAction; scriptA
...
+          const startColumn = editorModel.getPositionAt(textIndex + 1 + lengthUntilOpeningParenthesis).column;
+          const endLineNumber = editorModel.getPositionAt(textIndex + text.length).lineNumber;
+          const endColumn = editorModel.getPositionAt(textIndex + text.length - 1).column;
+
+          const textrange = {
+            startLineNumber: startLineNumber,
+            startColumn: startColumn,
+            endLineNumber: endLineNumber,
+            endColumn: endColumn
+          };
```
ivy-lli, line 66: shorten this code by build directly the object:
```ts
const textrange = {
  startLineNumber: editorModel.getPositionAt(textIndex).lineNumber,
  ...
};
```

packages/editor/src/components/widgets/code-editor/useCodeEditor.ts
```diff
@@ -39,6 +39,38 @@ export const useMonacoEditor = (options?: { modifyAction?: ModifyAction; scriptA
     } else {
       const text = value.length > 0 && options?.modifyAction ? options.modifyAction(value) : value;
       editor.executeEdits('browser', [{ range: selection, text, forceMoveMarkers: true }]);
+      if (type === 'func') {
```
ivy-lli, line 42: this will no longer work if the other pr is merged right?

packages/editor/src/components/browser/FunctionBrowser.tsx
```diff
@@ -1,20 +1,153 @@
...
+    const type = parentRow
+      ? parentRow.original.returnType.packageName + '.' + parentRow.original.returnType.simpleName
+      : selectedRow.original.returnType.packageName + '.' + selectedRow.original.returnType.simpleName;
+    setType(type);
+    setMethod(selectedRow.original.name);
+    setParamTypes(paramTypesArray);
+    setSelectedFunctionDoc(doc);
+    setShowHelper(true);
+    props.onChange(selectedFunction);
+  }, [doc, props, rowSelection, table]);
```
ivy-lli, line 125: I have no idea what is happening here...
Any maybe you could gain also more speed / performance if you can split this into smaller parts...
This use effect is now triggered if the doc, the props, a rowSelection or the whole table changes...

packages/editor/src/components/widgets/code-editor/useCodeEditor.ts
ivy-edp: I need the timeout because otherwise the focus is not set. Without the timeout the focus is on the button to open the browser after the "apply" and not on the monaco editor itself. It is a similar behavior as with the focushanldig of the search input, where I also had to work with a timeout to ensure that the corresponding command is not overwritten.

packages/editor/src/components/widgets/code-editor/useCodeEditor.ts
ivy-edp: Yes, that's right. i'll merge the function-browser first and then adapt the other pr accordingly. i didn't realize that i needed the type after all.

ivy-lli: not necessarily, you could aso invent another paremeter in this browser value type, e.g. like selectFunctionParam

packages/editor/src/components/widgets/code-editor/useCodeEditor.ts
ivy-lli: I think we could definitly listen on ended transitions and only then set focus stuff, e.g. like it is described here: https://developer.mozilla.org/en-US/docs/Web/API/Element/transitionend_event 
or the same for animation events: https://developer.mozilla.org/en-US/docs/Web/API/Element/animationend_event

packages/editor/src/components/browser/FunctionBrowser.tsx
ivy-edp: Yes, you are right, it's a bit confusing and the useEffect currently reacts to many changes. However, I wouldn't know how to split it up better because, when selecting a row, I need to perform all these tasks. 

In general, this useEffect is meant to control the behavior when a row is selected. When a row is selected, I need to first traverse back through the tree (using the getParentNames method) to obtain the complete name of the method. All the subsequent set... operations are required for calling the JavaDoc/helpertext. I've now tried to structure the code a bit better. If it's okay with you, I would merge the code like this and in a new PR, attempt to improve it (if possible). I'll have to modify the code again anyway to optimize the search.
grafik

ivy-lli: Yes I know that this is used for the row selection, but it runs also if the tree changes, and the tree changes e.g. when you search for something. 
But yes, merge it first then we can improve it :)

ivy-lli (approved):

PR #460, opened 2023-12-20
XIVY-12549 Fix inserted string for functionbrowser

ivy-edp:
Fix inserted string to be this:
ivy.cal.get(String).getBusinessDuration(DateTime, DateTime)
instead of this:
ivy.cal.get.getBusinessDuration(DateTime, DateTime)

I have adjusted the automatic marking accordingly so that only the parameters of the first function are marked.


ivy-lli (approved):
This is also something that you easily could extract into a standalone helper function and add some tests to it.

packages/editor/src/components/browser/functionBrowser/FunctionBrowser.tsx
```diff
@@ -1,159 +1,145 @@
...
+      parentRow
+        ? parentRow.original.returnType.packageName + '.' + parentRow.original.returnType.simpleName
+        : selectedRow.original.returnType.packageName + '.' + selectedRow.original.returnType.simpleName
+    );
+    setMethod(selectedRow.original.name);
+    setParamTypes(selectedRow.original.params.map(param => param.type));
+    //setup Helpertext
+    setSelectedFunctionDoc(doc);
+    setShowHelper(true);
+  }, [doc, props, rowSelection, table]);
```
ivy-edp, line 117: Oh, I think because I moved each browser to its own folder, it no longer shows the individual changes. But actually, I only removed the getParentNames method here in the function browser, as I have outsourced it to the functionBrowser-utils.

packages/editor/src/components/browser/Browser.tsx
```diff
@@ -4,15 +4,14 @@ import { Button, TabContent, TabList, TabRoot } from '../widgets';
...
-import { useAttributeBrowser } from './AttributeBrowser';
 import type { BrowserType, UseBrowserReturnValue } from './useBrowser';
-import { useFuncBrowser } from './FunctionBrowser';
-import { useTypeBrowser } from './TypeBrowser';
-import { useTableColBrowser } from './TableColBrowser';
-import { useSqlOpBrowser } from './SqlOperationBrowser';
-import { useCatPathChooserBrowser } from './CatPathChooser';
-import type { CmsOptions } from './CmsBrowser';
-import { useCmsBrowser } from './CmsBrowser';
+import { useAttributeBrowser } from './attributeBrowser/AttributeBrowser';
```
ivy-lli, line 8: we are already in the browser folder. Why duplicate it here?
just call them what browser is inside. `attributeBrowser` -> `attribute`.
And of course for the other too!

packages/editor/src/components/browser/Browser.tsx
```diff
@@ -4,15 +4,14 @@ import { Button, TabContent, TabList, TabRoot } from '../widgets';
...
+import { useCmsBrowser, type CmsOptions } from './cmsBrowser/CmsBrowser';
+import { useFuncBrowser } from './functionBrowser/FunctionBrowser';
+import { useTypeBrowser } from './typeBrowser/TypeBrowser';
+import { useCatPathChooserBrowser } from './catPathChooser/CatPathChooser';
```
ivy-lli, line 12: 'catPathChosser/CatPathChooser' ->' categorie/CatPathChooser'

packages/editor/src/components/browser/functionBrowser/FunctionBrowser.tsx
ivy-lli: Don't understand what you mean here...

packages/editor/src/components/browser/functionBrowser/functionBrowser-utils.ts
ivy-lli, line 1: Also this file name...
Why always duplicate this browser name?
browser/functionBrowser/functionBrowser-utils.ts why not simply call it what it's including -> browser/function/parent-name.ts

packages/editor/src/components/browser/typeBrowser/typeBrowser-utils.ts
ivy-lli, line 1: i would also rename this file -> /browser/type/cursor-value.ts

ivy-lli (approved):
Looks good, but I highly recommend thinking about your naming strategy...

PR #462, opened 2023-12-21
XIVY-12992 Fix ParameterTable with browser

ivy-edp:
When adding a type from the browser using the ParameterTable for input parameters, only [Object object] appeared instead of the actual type. This occurred because I am now working not only with strings but also with the BrowserValue object. This adjustment ensures that the table functions smoothly again.


ivy-lli (approved):

ivy-lli:
also merge down to release 11.2?

ivy-edp:
Yes, I will Cherry-Pick it afterwards.

PR #466, opened 2023-12-22
XIVY-12992 Fix EditableCell with Browser

ivy-edp:


ivy-lli (approved):
Same as reto said tomorrow.
If the release 11.2 pr is more or less the same as on the master, you don't need to add me for approval...

PR #469, opened 2024-01-04
XIVY-12797 Impl. of expandable Code Editor

ivy-edp:
I have now created the expandable code editor. I have already integrated it into the StartPart and TaskPart. However, before incorporating it into all other locations (approximately 7 additional places where the ScriptArea is used), I would like to hear your opinion on whether the integration in these two parts is okay. Even though we discussed earlier that I should leave the ScriptArea component as it is and not add additional parameters and if/else statements, it might now make sense to incorporate it directly there. This way, I could avoid code repetition in too many places. Additionally, I believe that the ScriptArea is only used in CodeParts, so the MaximizedCodeEditorBrowser always needs to be integrated with a ScriptArea. However, I might be mistaken, and it could be better to incorporate everything as shown in the StartPart and TaskPart, keeping the ScriptArea doing only what it was originally intended for. What do you think?

grafik
grafik


packages/editor/src/components/browser/MaximizedCodeEditorBrowser.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+  useEffect(() => {
+    const handleResize = () => {
+      setCodeEditorHeight(window.innerHeight - 150);
+    };
+    window.addEventListener('resize', handleResize);
+    return () => {
+      window.removeEventListener('resize', handleResize);
+    };
+  }, []);
```
ivy-lli, line 43: pretty sure this is not needed

packages/editor/src/components/browser/MaximizedCodeEditorBrowser.tsx
```diff
@@ -0,0 +1,84 @@
...
+              value={editorValue}
+              onChange={applyEditor}
+              context={{ location }}
+              height={codeEditorHeight}
+              onMountFuncs={[setEditor]}
+              options={MAXIMIZED_MONACO_OPTIONS}
+            />
+          </div>
+          <Browser {...browser} types={browsers} accept={modifyEditor} location={location} initSearchFilter={getMonacoSelection} />
+        </div>
```
ivy-lli, line 60: make own component

packages/editor/src/components/browser/useMaximizedCodeEditor.ts
```diff
@@ -0,0 +1,19 @@
...
+
+  const maximizeCode: FieldsetControl = {
+    label: 'Fullsize Code Editor',
+    icon: IvyIcons.ArrowsMaximize,
+    action: () => {
+      setIsMaximizedCodeEditorOpen(true);
+    }
+  };
+
+  return { isMaximizedCodeEditorOpen, maximizeCode, setIsMaximizedCodeEditorOpen };
```
ivy-lli, line 16:
```
return { openState: {isMaximizedCodeEditorOpen, setIsMaximizedCodeEditorOpen}, maximizeCode };
```

ivy-lli (changes requested):

packages/editor/src/components/browser/MaximizedCodeEditorBrowser.tsx
```diff
@@ -0,0 +1,41 @@
...
+      icon: IvyIcons.StartProgram
+    }
+  ];
+
+  return (
+    <>
+      <Dialog open={open} onOpenChange={onOpenChange}>
+        <BrowserBody activeTab={'maxCode'} open={open} tabs={tabs} />
+      </Dialog>
+    </>
```
ivy-lli, line 37: fragment not needed

packages/editor/src/components/widgets/code-editor/ScriptArea.tsx
```diff
@@ -4,13 +4,30 @@ import ResizableCodeEditor from './ResizableCodeEditor';
 import { Browser, useBrowser } from '../../../components/browser';
 import { useMonacoEditor } from './useCodeEditor';
 import { usePath } from '../../../context';
+import MaximizedCodeEditorBrowser from '../../browser/MaximizedCodeEditorBrowser';
 
-const ScriptArea = (props: CodeEditorAreaProps) => {
+type ScriptAreaProps = CodeEditorAreaProps & {
+  openState: {
```
ivy-lli, line 10: maybe maximizeState or something like that is a bit clearer for a user of the ScriptArea...

packages/editor/src/components/widgets/headlineControls/HeadlineControls.tsx
```diff
@@ -0,0 +1,30 @@
+import './HeadlineControls.css';
+import Button from '../button/Button';
+import type { FieldsetControl } from '../fieldset';
+
+type HeadlineControlsProps = {
+  controls: FieldsetControl[];
+  open?: boolean;
```
ivy-lli, line 7: the open has nothing to do with this component IMO. This is only a special case for the collapsible, that if it is open the controls are shown. So just handle it there.

packages/editor/src/monaco/monaco-editor-util.ts
```diff
@@ -23,6 +23,17 @@ export const MONACO_OPTIONS: monaco.editor.IStandaloneEditorConstructionOptions
...
 };
 
+export const MAXIMIZED_MONACO_OPTIONS: monaco.editor.IStandaloneEditorConstructionOptions = {
+  ...MONACO_OPTIONS,
+  lineNumbers: 'on',
+  folding: true,
+  renderLineHighlight: 'gutter',
+  cursorStyle: 'line',
+  cursorBlinking: 'blink',
+  contextmenu: false, // Disabling context menu
```
ivy-lli, line 33: already in the default MONACO_OPTIONS

packages/editor/src/monaco/monaco-editor-util.ts
```diff
@@ -23,6 +23,17 @@ export const MONACO_OPTIONS: monaco.editor.IStandaloneEditorConstructionOptions
   fixedOverflowWidgets: true
 };
 
+export const MAXIMIZED_MONACO_OPTIONS: monaco.editor.IStandaloneEditorConstructionOptions = {
+  ...MONACO_OPTIONS,
+  lineNumbers: 'on',
+  folding: true,
+  renderLineHighlight: 'gutter',
+  cursorStyle: 'line',
+  cursorBlinking: 'blink',
```
ivy-lli, line 32: isn't this the default?

packages/editor/src/components/widgets/collapsible/Collapsible.tsx
```diff
@@ -22,14 +30,15 @@ const Collapsible = ({ label, defaultOpen, message, children, autoClosable }: Co
   }, [autoClosable, defaultOpen]);
   return (
     <CollapsibleRoot className='collapsible-root' open={open} onOpenChange={setOpen}>
-      <div className='collapsible-header'>
+      <div className={`collapsible-header${!open ? '-closed' : ''}`}>
```
ivy-lli, line 33: I wouldn't change the whole class. Makes it only much harder to test IMO.
Better don't change the default component class, and handle such state stuff in a second class or in a data attribute e.g. `data-state={open ? 'open' : 'closed'}`

And if you do such things, I can only speak for myself, but IMO negative stuff is much harder to read. So even if you don't change this to the attribute style I would write it like `open ? '' : '-closed'`

ivy-lli (changes requested):

ivy-lli:
a lot better IMO. Some small stuff to improve :)

packages/editor/src/monaco/monaco-editor-util.ts
ivy-edp: yes, you're right 👍

packages/editor/src/components/widgets/collapsible/Collapsible.css
```diff
@@ -1,10 +1,10 @@
-.collapsible-root .collapsible-header {
+.collapsible-root .collapsible-header:has([data-state='open']) {
```
ivy-lli, line 1: `:has` is not needed here
just do `.collapse-header[data-state='open']`

And I think you should have a default style without a data-state and then the style for a specific data state e.g. 
```css
.collapsible-root .collapsible-header {
  display: flex;
}
.collapsible-root .collapsible-header[data-state='open'] {
  justify-content: space-between;
  align-items: center;
}
```

ivy-lli: Ah if you want to check the child state, yes then has is needed.
```css
.collapsible-root .collapsible-header {
  display: flex;
}
.collapsible-root .collapsible-header:has([data-state='open']) {
  justify-content: space-between;
  align-items: center;
}
```

ivy-lli (approved):

ivy-lli:
Cool 👍

PR #470, opened 2024-01-05
XIVY-12797 Add open with f2 and adopt cursorPosition/selection feature

ivy-edp:
I have now implemented two features to the maximized code editor:
The maximized code editor can now be opened using the "F2" key.
The cursor position or selection from the minimized code editor is also transferred to the maximized code editor. Unfortunately, what doesn't work as well now is directly closing the browser again with the "Esc" key. To make this work, you have to first click outside the Monaco editor to lose focus and then it works.

What I have not implemented (yet) is transferring the cursor position and selection from the maximized code editor to the minimized code editor. However, I'm not sure if this is really necessary, as I don't expect someone to want to edit in the minimized editor immediately after working in the maximized code editor, especially right after clicking "apply." What do you think?

Lastly, I attempted to implement the entire "focusWithin" aspect in the ScriptArea similar to the ScriptInput. Unfortunately, the editor still behaves strangely when I click "apply" so i reverted these changes. I don't quite understand why but we can take a look at it together if you'd like. However, I now believe that it's not a bad idea to always have the Monaco Editor present, as otherwise, the cursor position and selection feature would no longer work or would become much more complicated to implement (I think).

### DIFF
--- a/packages/editor/src/components/browser/maximizedCodeEditor/MaximizedCodeEditor.tsx
+++ b/packages/editor/src/components/browser/maximizedCodeEditor/MaximizedCodeEditor.tsx
@@ -1,20 +1,38 @@
 import './MaximizedCodeEditor.css';
+import * as monaco from 'monaco-editor/esm/vs/editor/editor.api';
 import { useBrowser, type BrowserType } from '../useBrowser';
-import { useMonacoEditor } from '../../widgets/code-editor/useCodeEditor';
+import { monacoAutoFocus, useMonacoEditor } from '../../widgets/code-editor/useCodeEditor';
 import CodeEditor from '../../widgets/code-editor/CodeEditor';
 import Browser from '../Browser';
 import { MAXIMIZED_MONACO_OPTIONS } from '../../../monaco/monaco-editor-util';
 
-type MaximizedCodeEditorProps = {
+export type MaximizedCodeEditorProps = {
   editorValue: string;
   location: string;
   browsers: BrowserType[];
   applyEditor: (change: string) => void;
+  selectionRange: monaco.IRange | null;
+  keyActions?: {
+    escape?: () => void;
+  };
 };
 
-const MaximizedCodeEditor = ({ editorValue, location, browsers, applyEditor }: MaximizedCodeEditorProps) => {
-  const { setEditor, modifyEditor, getMonacoSelection } = useMonacoEditor({ scriptAreaDatatypeEditor: true });
+const MaximizedCodeEditor = ({ editorValue, location, browsers, applyEditor, selectionRange, keyActions }: MaximizedCodeEditorProps) => {
+  const { setEditor, modifyEditor, getMonacoSelection } = useMonacoEditor();
   const browser = useBrowser();
+
+  const setSelection = (editor: monaco.editor.IStandaloneCodeEditor) => {
+    if (selectionRange !== null) {
+      editor.setSelection(selectionRange);
+    }
+  };
+  const keyActionMountFunc = (editor: monaco.editor.IStandaloneCodeEditor) => {
+    editor.addCommand(monaco.KeyCode.Escape, () => {
+      if (keyActions?.escape) {
+        keyActions.escape();
+      }
+    });
+  };
 
   return (
     <div className='maximized-script-area'>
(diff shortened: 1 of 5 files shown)


packages/editor/src/components/widgets/code-editor/ScriptArea.tsx
```diff
@@ -15,9 +17,22 @@ type ScriptAreaProps = CodeEditorAreaProps & {
...
+      if (event.key === 'F2' && isFocusWithin) {
+        event.preventDefault();
+        props.maximizeState.setIsMaximizedCodeEditorOpen(true);
+      }
+    };
+    document.addEventListener('keydown', handleKeyDown);
+    return () => {
+      document.removeEventListener('keydown', handleKeyDown);
+    };
+  }, [isFocusWithin, props.maximizeState]);
```
ivy-lli, line 35: did you checkout the other code editors? they already have a concept for keybindings... I would solve it more like those

packages/editor/src/components/widgets/code-editor/useCodeEditor.ts
```diff
@@ -13,9 +13,29 @@ export const monacoAutoFocus = (editor: monaco.editor.IStandaloneCodeEditor) =>
...
+        editor.setPosition(options.maximizedCodeEditor.cursorPosition);
+      }
+      if (options.maximizedCodeEditor.selectionRange !== null) {
+        editor.setSelection(options.maximizedCodeEditor.selectionRange);
+      }
+      setTimeout(() => {
+        editor.focus();
+      }, 500);
+    }
+  }
```
ivy-lli, line 37: I don't fully understand why this inside this hook? There is already a concept on run some functions against the monaco editor `onMountFuncs` where you can add an array of functions which are run when the editor is mounted...
Handle this focus in a timeout screems to me 'I will behave very strange or break in the future'. And by the way there is already a monacoAutoFocus function which you can simply add to the onMountFuncs...
I would expect this selection stuff to be also something like that.

packages/editor/src/components/browser/maximizedCodeEditor/MaximizedCodeEditor.tsx
```diff
@@ -1,19 +1,29 @@
...
 import { MAXIMIZED_MONACO_OPTIONS } from '../../../monaco/monaco-editor-util';
 
-type MaximizedCodeEditorProps = {
+export type MaximizedCodeEditorProps = {
   editorValue: string;
   location: string;
   browsers: BrowserType[];
   applyEditor: (change: string) => void;
+  cursorPosition: { lineNumber: number; column: number } | null;
+  selectionRange: monaco.IRange | null;
```
ivy-lli, line 15: isn't this the same? if I have a selection the cursor should be at the end of this selection and if the selection is nothing then you have still the position in there (endlinenumber, endcolumn)?

ivy-lli (changes requested):
first you don't need to write a full story to me in this reviews, a bullet point list or something like that is more than enough for me. 

second you don't have to reinvent everything. For me it helps when I come to a problem quickly think if this exist somewhere else and is maybe already fixed there. e.g. the monaco in the popup in a mapping table: you get autofocus if it opens, you can press escape or tab to make the popover close. sounds very similar to your problems IMO. 
We should always try to solve problems similar, so we can remove duplicated code and if someone new gets to the code he doesn't need to learn 4 different ways how something is solved. 
And by the way more code means always more bugs: https://stackoverflow.com/a/56043694

ivy-lli (commented):
In general a very good improvement or what do you think?
The only thing I'm not sure about is this new forwardref which you only need for the newly added method to quickly escape the full screen editor right?

For me, forwardrefs are very complicated to read and should only be used with good reasons. Of course, it is a cool feature, but one of those which are not worth of this complication IMO. And if this escape works here then it should work on every browser. Or does it already work there?
Then we should implement it here too yes, but I'm pretty sure that there are better ways to handle this (e.g. simply put the method which lets you set the browser open state, this would be much clearer what this is for).

ivy-edp:
In general, it is the case that this ESC functionality is already present in all other browsers and works without any issues. Unfortunately, this is not the case with the maximized code editor, as I assume that this is related to having the focus directly in the Monaco editor when opened, causing the loss of the pre-existing ESC functionality, which needs to be specifically addressed here. Regarding the "forwarderef": Okay, yes, you are right. I have now solved it as you suggested, passing the onChangeOpen function instead of the "forwarderef." This also works without any problems👍.

ivy-lli (approved):
Cool thanks for the improvement 🥇

PR #472, opened 2024-01-08
XIVY-13156 Imrpove Improvements for function browser search

ivy-edp:
Improvement of the function browser search:
- The virtualizer is utilized.
- Only a maximum of 100 entries are displayed at a time. Additional entries (in increments of 100) can be loaded at the bottom.
- The data is loaded alphabetically, and the root is always expanded.

### DIFF
--- a/packages/editor/src/components/browser/function/FunctionBrowser.tsx
+++ b/packages/editor/src/components/browser/function/FunctionBrowser.tsx
@@ -1,5 +1,5 @@
-import { useEffect, useMemo, useState } from 'react';
-import { ExpandableCell, SelectRow, Table, TableCell } from '../../widgets';
+import { useEffect, useMemo, useRef, useState } from 'react';
+import { ExpandableCell, SelectRow, Table, TableCell, TableFooter, TableShowMore } from '../../widgets';
 import type { UseBrowserImplReturnValue } from '../useBrowser';
 import type { Function } from '@axonivy/inscription-protocol';
 import {
@@ -12,6 +12,7 @@ import {
   getFilteredRowModel,
   flexRender
 } from '@tanstack/react-table';
+import { useVirtualizer } from '@tanstack/react-virtual';
 import { IvyIcons } from '@axonivy/editor-icons';
 import type { BrowserValue } from '../Browser';
 import { useEditorContext, useMeta } from '../../../context';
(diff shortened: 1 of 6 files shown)


packages/editor/src/components/browser/function/FunctionBrowser.tsx
```diff
@@ -34,14 +36,38 @@ const FunctionBrowser = (props: { value: string; onChange: (value: BrowserValue)
...
   const { data: doc } = useMeta('meta/scripting/apiDoc', { context, method, paramTypes, type }, '');
 
+  const [rowNumber, setRowNumber] = useState(100);
+
   const [selectedFunctionDoc, setSelectedFunctionDoc] = useState('');
 
   const [showHelper, setShowHelper] = useState<boolean>(false);
 
+  useEffect(() => {
+    console.log('Tree data:', tree);
```
ivy-lli, line 50: remove log

packages/editor/src/components/browser/function/FunctionBrowser.tsx
```diff
@@ -123,19 +159,34 @@ const FunctionBrowser = (props: { value: string; onChange: (value: BrowserValue)
...
         <tbody>
-          {table.getRowModel().rows.map(row => (
-            <SelectRow key={row.id} row={row} onDoubleClick={props.onDoubleClick}>
-              {row.getVisibleCells().map(cell => (
-                <TableCell key={cell.id}>{flexRender(cell.column.columnDef.cell, cell.getContext())}</TableCell>
-              ))}
-            </SelectRow>
-          ))}
+          {virtualizer.getVirtualItems().map(virtualRow => {
+            const row = rows[virtualRow.index] as Row<Function>;
```
ivy-lli, line 171: why is this `as` needed?

packages/editor/src/components/browser/function/FunctionBrowser.tsx
```diff
@@ -123,19 +159,34 @@ const FunctionBrowser = (props: { value: string; onChange: (value: BrowserValue)
...
+                ))}
+              </SelectRow>
+            );
+          })}
         </tbody>
+        {rowNumber < rows.length && (
+          <TableFooter>
+            <TableShowMore
+              colSpan={4}
+              showMore={() => setRowNumber(rowNumber + 100)}
```
ivy-lli, line 185: normally if you use the old value to calculate the new one the method `setRowNumber(old => old + 100);` fits better.

packages/editor/src/components/widgets/table/footer/TableFooter.tsx
```diff
@@ -18,3 +18,16 @@ export const TableAddRow = (props: { colSpan: number; addRow: () => void }) => {
     </tr>
   );
 };
+
+export const TableShowMore = (props: { colSpan: number; showMore: () => void; helpertext: string }) => {
+  const readonly = useReadonly();
```
ivy-lli, line 23: this shouldn't be needed, because in the readonly mode you either should be able to load more rows (why should this not be allowed?) or the browser shouldn't be opeable at all...

packages/editor/src/components/widgets/table/footer/TableFooter.tsx
```diff
@@ -18,3 +18,16 @@ export const TableAddRow = (props: { colSpan: number; addRow: () => void }) => {
...
+  return (
+    <tr>
+      <th colSpan={props.colSpan} className='show-more-rows'>
+        <Button rotate={90} onClick={props.showMore} disabled={readonly} aria-label='Show more rows'>
+          ...{props.helpertext}
```
ivy-lli, line 28: why do you not use the `dots` ivy icon?

ivy-lli (approved):
Looks good in general, still some things you could improve. But you can merge after that

packages/editor/src/components/widgets/table/footer/TableFooter.tsx
ivy-edp: Yes, I could do that, but I need to include more CSS magic for the icon to look the same. Otherwise, it would look something like this:
grafik
or this:
grafik

and without the icon it just looks like this:
grafik

what do you prefer?

ivy-lli: I had the second in mind. Not sure what you mean with a lot more css magic, but if it isn't worth (in your opinion), I would just show the text without the dots or any icon.

PR #476, opened 2024-01-09
XIVY-12797 Remove timeout and replace with eventlistener

ivy-edp:


ivy-lli (approved):
Looks like the right direction. Two things:
First I think you can easily create a shared function for this where you can add a function as parameter which is called when the dialog animation is ended.

Second: yes the document querySelector should work here, but I think this is exactly the point where a ref should be used.

PR #484, opened 2024-01-17
XIVY-13143 add process kind to newProcess action

ivy-lmu:
in order to trigger the right vs code command for process creation


ivy-lli (commented):
I know that you need this at the moment, but later maybe this should be solved differently...
But now this label would be more or less hardcoded without any type safety. If someone changes this (maybe accidentally) this will no longer work in the vscode extension.

I think it would make more sense to create a new action in the backend which then is known via protocol generation... like the html dialog create action

ivy-lli:
maybe this is automatically another problem, when you think about how the new signature of the created process is added to the element?

PR #486, opened 2024-01-19
XIVY-12550 Add CatPath Chooser

ivy-edp:
Implemented the Category Path Chooser with mock tests
grafik

### DIFF
--- a/packages/editor/src/components/browser/categorie/CatPathChooser.tsx
+++ b/packages/editor/src/components/browser/categorie/CatPathChooser.tsx
@@ -1,21 +1,157 @@
-import { useState } from 'react';
-import { Input } from '../../widgets';
+import { useEffect, useMemo, useState } from 'react';
+import { ExpandableCell, SelectRow, Table, TableCell } from '../../widgets';
 import type { UseBrowserImplReturnValue } from '../useBrowser';
 import { IvyIcons } from '@axonivy/editor-icons';
 import type { BrowserValue } from '../Browser';
+import { useEditorContext, useMeta } from '../../../context';
+import {
+  useReactTable,
+  type ColumnDef,
+  type RowSelectionState,
+  getCoreRowModel,
+  getFilteredRowModel,
+  flexRender
+} from '@tanstack/react-table';
+import type { CategoryPathMeta, Type1 } from '@axonivy/inscription-protocol';
 export const CAT_PATH_CHOOSER_BROWSER_ID = 'catPath' as const;
 
-export const useCatPathChooserBrowser = (): UseBrowserImplReturnValue => {
-  const [value, setValue] = useState<BrowserValue>({ cursorValue: 'category path chooser' });
+export const useCatPathChooserBrowser = (onDoubleClick: () => void, location: string): UseBrowserImplReturnValue => {
+  const [value, setValue] = useState<BrowserValue>({ cursorValue: '' });
+
   return {
     id: CAT_PATH_CHOOSER_BROWSER_ID,
     name: 'Category Path Chooser',
-    content: <CatPathChooserBrowser value={value.cursorValue} onChange={(change: string) => setValue({ cursorValue: change })} />,
+    content: <CatPathChooserBrowser value={value.cursorValue} onChange={setValue} location={location} onDoubleClick={onDoubleClick} />,
     accept: () => value,
     icon: IvyIcons.Label
   };
 };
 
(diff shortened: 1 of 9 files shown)


packages/editor/src/components/browser/categorie/CatPathChooser.tsx
```diff
@@ -1,21 +1,157 @@
...
+    const location = props.location.toLowerCase();
+    switch (true) {
+      case location.includes('request') || location.includes('start'):
+        return 'START';
+      case location.includes('case'):
+        return 'CASE';
+      case location.includes('task'):
+        return 'TASK';
+      default:
+        throw new Error('Unknown type');
```
ivy-lli, line 48: are you sure about this?
If this happens maybe the whole UI is broken. Is this what we want?
Or did you try it?
I think we can simply fire a request with an unknown location and the backend should then simply return nothing or something like this...

packages/editor/src/components/widgets/code-editor/useCodeEditor.ts
```diff
@@ -38,8 +38,14 @@ export const useMonacoEditor = (options?: { modifyAction?: ModifyAction }) => {
...
     } else {
-      const text = value.cursorValue.length > 0 && options?.modifyAction ? options.modifyAction(value.cursorValue) : value.cursorValue;
+      const text =
+        value.cursorValue.length > 0 && options?.modifyAction && type !== 'catPath'
+          ? options.modifyAction(value.cursorValue)
+          : value.cursorValue;
       editor.executeEdits('browser', [{ range: selection, text, forceMoveMarkers: true }]);
+      if (type === 'catPath') {
+        editor.setValue(text);
+      }
```
ivy-lli, line 48: arr, a little bit ugly, but I think I understand why it is needed...
Maybe a sign that we should somehow change this or at last refactor this code a bit into smaller readable functions...

packages/protocol/src/data/inscription.ts
```diff
@@ -6,7 +6,8 @@
  * and run json-schema-to-typescript to regenerate this file.
  */
 
-export type PID = string
+export type Type1 = ("START" | "TASK" | "CASE")
```
ivy-lli, line 9: ugly name... can't we improve this @ivy-rew?

packages/protocol/src/data/inscription.ts
```diff
@@ -21,7 +22,7 @@ export type HttpMethod = "GET" | "POST" | "PUT" | "HEAD" | "DELETE" | "PATCH" |
 export type InputType = "ENTITY" | "FORM" | "RAW";
 export type WsAuth = "NONE" | "WS_SECURITY" | "HTTP_BASIC";
 export type Severity = "INFO" | "WARNING" | "ERROR";
-export type Type = "START" | "INTERMEDIATE" | "ACTIVITY";
+export type Type2 = "START" | "INTERMEDIATE" | "ACTIVITY";
```
ivy-lli, line 25: ugly name... can't we improve this @ivy-rew?

ivy-lli (commented):
Looks good.
I would think about throwing an error -> this will not help any user I think.

And I'm not sure if the category browser should be a tree... Of course the macros will not be possible, but all other we could easily render as tree.
And I think I would not call this browser "Category Path chooser", only "Category".

packages/protocol/src/data/inscription.ts
ivy-rew: yes indeed; we should change that type-info in the backend ... by renaming our new object to "CategoryType"

ivy-edp: yes, I can do that👍

packages/protocol/src/data/inscription.ts
```diff
@@ -22,7 +22,7 @@ export type HttpMethod = "GET" | "POST" | "PUT" | "HEAD" | "DELETE" | "PATCH" |
 export type InputType = "ENTITY" | "FORM" | "RAW";
 export type WsAuth = "NONE" | "WS_SECURITY" | "HTTP_BASIC";
 export type Severity = "INFO" | "WARNING" | "ERROR";
-export type Type2 = "START" | "INTERMEDIATE" | "ACTIVITY";
+export type Type = "START" | "INTERMEDIATE" | "ACTIVITY";
```
ivy-lli, line 25: I think a better name for this would be nice too, but not important for now

ivy-lli (approved):
Technically good now I think. But as discussed IMO this should be a combobox like the signal boundary has.

PR #492, opened 2024-01-29
XIVY-13253 Add custom Table use-hook and improve SelectableValidationRow

ivy-edp:
I created my own use-hook to reuse the table, as well as rowSelection and add/remove row functions. What do you think about that? Maybe the naming of the custom hook could be improved 😅 

Additionally, I have also improved the SelectableValidationRow as mentioned in the previous PR.

### DIFF
--- a/packages/editor/src/components/widgets/table/row/MessageRow.tsx
+++ b/packages/editor/src/components/widgets/table/row/MessageRow.tsx
@@ -3,24 +3,37 @@ import type { ComponentProps } from 'react';
 import { forwardRef } from 'react';
 import type { MessageTextProps } from '../../message/Message';
 import { MessageText } from '../../message/Message';
+import type { InscriptionValidation } from '@axonivy/inscription-protocol';
 
 export type MessageRowProps = MessageTextProps & ComponentProps<'tr'> & { colSpan: number };
 
-export const MessageRow = forwardRef<HTMLTableRowElement, MessageRowProps>(
+export const MessageRowWithTr = forwardRef<HTMLTableRowElement, MessageRowProps>(
   ({ message, children, className, colSpan, ...props }, forwardRef) => {
     return (
       <>
-        <tr ref={forwardRef} className={`row ${message ? `row-${message.severity.toLocaleLowerCase()}` : ''} ${className}`} {...props}>
+        <tr ref={forwardRef} className={styleMessageRow(message, className)} {...props}>
           {children}
         </tr>
-        {message && (
-          <tr className='row row-message'>
-            <td colSpan={colSpan}>
-              <MessageText message={message} />
-            </td>
-          </tr>
-        )}
+        <MessageRow message={message} colSpan={colSpan} />
       </>
     );
   }
 );
+
+export const styleMessageRow = (message?: Omit<InscriptionValidation, 'path'>, className?: string) => {
+  return `row ${message ? `row-${message.severity.toLocaleLowerCase()}` : ''} ${className}`;
+};
+
+export const MessageRow = (props: { message?: Omit<InscriptionValidation, 'path'>; colSpan?: number }) => {
+  return (
+    <>
+      {props.message && (
+        <tr className='row row-message'>
+          <td colSpan={props.colSpan ?? 2}>
+            <MessageText message={props.message} />
+          </td>
+        </tr>
+      )}
+    </>
+  );
+};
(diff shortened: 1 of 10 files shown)


packages/editor/src/components/widgets/table/row/MessageRow.tsx
```diff
@@ -4,15 +4,17 @@ import { forwardRef } from 'react';
...
     return (
       <>
-        <tr ref={forwardRef} className={`row ${message ? `row-${message.severity.toLocaleLowerCase()}` : ''} ${className}`} {...props}>
-          {children}
-        </tr>
+        {!selectable && (
+          <tr ref={forwardRef} className={`row ${message ? `row-${message.severity.toLocaleLowerCase()}` : ''} ${className}`} {...props}>
+            {children}
+          </tr>
+        )}
```
ivy-lli, line 17: not exaclty the way I would do it...
I would split this up into two components one for each row, which then can be reused everywhere else...
And still the select row misses the row-<severiy> class right? so it will not show a red border around the row if something is wrong

packages/editor/src/context/useResizableEditableTable.ts
```diff
@@ -0,0 +1,84 @@
+import type { ColumnDef, RowSelectionState, SortingState } from '@tanstack/react-table';
+import { getCoreRowModel, getSortedRowModel, useReactTable } from '@tanstack/react-table';
+
+import { useState, useEffect } from 'react';
+import { deepEqual } from '../utils/equals';
+interface UseResizableEditableTableProps<TData> {
+  data: TData[];
+  columns: ColumnDef<TData>[];
+  onChange: (change: TData[]) => void;
+  EMPTY_DATA_OBJECT: TData;
```
ivy-lli, line 10: I know we have the convention that constants are written full uppercase. But here it is a parameter, so this one should definitely be lowercase.

packages/editor/src/context/useResizableEditableTable.ts
```diff
@@ -0,0 +1,84 @@
...
+    newData.splice(index, 1);
+    if (newData.length === 0) {
+      setRowSelection({});
+    } else if (index === data.length - 1) {
+      setRowSelection({ [`${newData.length - 1}`]: true });
+    }
+    onChange(newData);
+  };
+
+  return { table, rowSelection, setRowSelection, addRow, removeRow };
```
ivy-lli, line 81: I would also return a function that has the state if the add row should be displayed: `showAddButton` or something the like.

ivy-lli: because then you don't have to do anything with this empty object outside of this object.

packages/editor/src/context/useResizableEditableTable.ts
ivy-lli, line 1: did you make a context here? or another question, do you know what a context is?

ivy-lli (changes requested):
Looks very promissing.

Some comments to improve:

packages/editor/src/context/useResizableEditableTable.ts
ivy-edp: No and no 😅 But if I have to make a context here, I can certainly try to read into this topic.

packages/editor/src/components/widgets/table/row/MessageRow.tsx
ivy-edp: I'm not entirely sure if I've understood your idea correctly. Do you want to modify the MessageRow component so that it only consists of the message row, and create a new second component that reflects the select row, but additionally includes the CSS class of MessageRow? And than reuse both components within the SelectableValidationRow?

For now, I've adjusted the SelectRow to set the corresponding CSS class for the red border. Otherwise, we can briefly discuss how you envisioned it with the two components.

packages/editor/src/context/useResizableEditableTable.ts
ivy-lli: Yes then you should definitely read into this topic! 
Because no, you do no context here, so I don't understand why this hook is in the context folder.

packages/editor/src/components/widgets/table/row/MessageRow.tsx
ivy-lli: let's have a look at it together tomorrow

packages/editor/src/components/widgets/table/row/SelectRow.tsx
```diff
@@ -1,18 +1,26 @@
...
-import type { ReactNode } from 'react';
+import type { ComponentProps, ReactNode } from 'react';
 
-export type SelectRowProps<TData> = {
+export type SelectRowProps<TData> = ComponentProps<'tr'> & {
   row: Row<TData>;
   children: ReactNode;
   title?: string;
   isNotSelectable?: boolean;
+  additionalClassName?: string;
```
ivy-lli, line 10: not needed any longer right? as you reuse the already existing className prop

ivy-lli (approved):
Great!

PR #493, opened 2024-01-30
Fix IT for General/Start/Result

ivy-edp:
Currently, some IT are failing. As I have observed, the remove functionality does not work idealy with WebKit and Firefox. If there is only one row present and it is removed via a click on the remove button, the cleanup functionality is not always triggered. Therefore, I now additionally click the header to ensure that the empty rows are removed.


ivy-lli (commented):
Is it not triggered or does this problem occur because webkit and firefox are slower?
Another way would be to check after the remove button click that the row is really removed...

PR #497, opened 2024-02-06
XIVY-13253: Adjust all other Tables

ivy-edp:
I've now adjusted all other tables: RestForm, PropertyTable, ConditionTable, TableSort, MailAttachment, and TableFields. 

Unfortunately, I couldn't implement automatic row addition and deletion functionality for tables with ComboCells due to encountering use-hook loops. Despite investing some time investigating the issue, I couldn't find a solution. Therefore, I decided to remove it, considering that the Rest-Acctivity might be revisited anyway. Additionally, there are still some tests (UT and IT) that I haven't managed to fix yet. I'll skip them for this pull request and address them later.

### DIFF
--- a/integrations/standalone/tests/pageobjects/Section.ts
+++ b/integrations/standalone/tests/pageobjects/Section.ts
@@ -33,6 +33,7 @@ export class Section extends Composite {
   }
 
   private static toggleButtonLocatorInternal(page: Page, label: string) {
-    return page.locator(`.collapsible-trigger:has-text("${label}")`);
+    const regexLabel = new RegExp(`^${label}$`);
+    return page.locator(`.collapsible-trigger`, { hasText: regexLabel });
   }
 }
--- a/packages/editor/src/components/parts/common/path/validation/ValidationRow.tsx
+++ b/packages/editor/src/components/parts/common/path/validation/ValidationRow.tsx
@@ -1,7 +1,7 @@
 import type { ReactNode } from 'react';
 import { mergePaths, usePath, useValidations } from '../../../../../context';
-import type { ReorderRowProps, SelectRowProps } from '../../../../widgets';
-import { MessageRow, MessageRowWithTr, ReorderRow, SelectRow, styleMessageRow } from '../../../../widgets';
+import type { MessageRowProps, ReorderRowProps, SelectRowProps } from '../../../../widgets';
+import { MessageRow, SelectRow, SelectableReorderRow, styleMessageRow } from '../../../../widgets';
 
 type ValidationRowProps = {
   rowPathSuffix: string | number;
@@ -15,15 +15,6 @@ const useValidationRow = (rowPathSuffix: string | number) => {
   const path = usePath();
   const rowPath = mergePaths(path, [rowPathSuffix]);
   return validations.find(val => val.path === rowPath);
-};
-
-export const ValidationRow = ({ rowPathSuffix, children, title, colSpan }: ValidationRowProps) => {
-  const message = useValidationRow(rowPathSuffix);
-  return (
-    <MessageRowWithTr message={message} title={title} colSpan={colSpan ? colSpan : 2}>
-      {children}
-    </MessageRowWithTr>
-  );
 };
 
 export const SelectableValidationRow = <TData extends object>({
@@ -45,13 +36,17 @@ export const SelectableValidationRow = <TData extends object>({
   );
 };
 
-type ValidationReorderRowProps = ValidationRowProps & ReorderRowProps;
+type ValidationReorderRowProps<TData> = ValidationRowProps & ReorderRowProps & SelectRowProps<TData> & MessageRowProps;
 
-export const ValidationReorderRow = ({ rowPathSuffix, children, ...props }: ValidationReorderRowProps) => {
+export const ValidationSelectableReorderRow = <TData extends object>({
+  rowPathSuffix,
+  children,
+  ...props
+}: ValidationReorderRowProps<TData>) => {
   const message = useValidationRow(rowPathSuffix);
   return (
-    <ReorderRow message={message} {...props}>
+    <SelectableReorderRow message={message} {...props}>
       {children}
-    </ReorderRow>
+    </SelectableReorderRow>
   );
 };
--- a/packages/editor/src/components/parts/common/table/useResizableEditableTable.ts
+++ b/packages/editor/src/components/parts/common/table/useResizableEditableTable.ts
@@ -1,8 +1,9 @@
 import type { ColumnDef, RowSelectionState, SortingState } from '@tanstack/react-table';
 import { getCoreRowModel, getSortedRowModel, useReactTable } from '@tanstack/react-table';
-
 import { useState, useEffect } from 'react';
 import { deepEqual } from '../../../../utils/equals';
+import type { FieldsetControl } from '../../../widgets';
+import { IvyIcons } from '@axonivy/editor-icons';
 
 interface UseResizableEditableTableProps<TData> {
   data: TData[];
@@ -83,7 +84,13 @@ const useResizableEditableTable = <TData>({ data, columns, onChange, emptyDataOb
     onChange(newData);
   };
 
-  return { table, rowSelection, setRowSelection, addRow, removeRow, showAddButton };
+  const removeRowAction: FieldsetControl = {
+    label: 'Remove row',
+    icon: IvyIcons.Trash,
+    action: () => removeRow(table.getRowModel().rowsById[Object.keys(rowSelection)[0]].index)
+  };
+
+  return { table, rowSelection, removeRowAction, setRowSelection, addRow, showAddButton };
 };
 
 export { useResizableEditableTable };
--- a/packages/editor/src/components/parts/condition/ConditionTable.tsx
+++ b/packages/editor/src/components/parts/condition/ConditionTable.tsx
@@ -1,29 +1,17 @@
-import { Button, ScriptCell, EditableCell, Table, TableCell, TableHeader } from '../../widgets';
-import { useCallback, useMemo } from 'react';
+import { Table, TableCell, TableHeader, Fieldset, ResizableHeader, ScriptCell, ReorderWrapperIcon, TextHeader } from '../../widgets';
+import { useCallback, useMemo, useState } from 'react';
 import { Condition } from './condition';
-import type { CellContext, ColumnDef } from '@tanstack/react-table';
-import { flexRender, getCoreRowModel, useReactTable } from '@tanstack/react-table';
+import type { ColumnDef, RowSelectionState, SortingState } from '@tanstack/react-table';
+import { flexRender, getCoreRowModel, getSortedRowModel, useReactTable } from '@tanstack/react-table';
 import { IvyIcons } from '@axonivy/editor-icons';
 import { IVY_SCRIPT_TYPES } from '@axonivy/inscription-protocol';
-import { ValidationReorderRow } from '../common';
+import { ValidationSelectableReorderRow } from '../common';
 
 const ConditionTypeCell = ({ condition }: { condition: Condition }) => {
   if (condition.target) {
     return <span>{`${condition.target.name}: ${condition.target.type.id}`}</span>;
   }
   return <span>⛔ {condition.fid}</span>;
-};
-
-const ConditionExpressionCell = ({ cell, removeCell }: { cell: CellContext<Condition, unknown>; removeCell: (id: string) => void }) => {
-  if (cell.row.original.target) {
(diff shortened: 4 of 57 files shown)


packages/editor/src/components/widgets/table/row/ReorderRow.tsx
```diff
@@ -37,22 +46,31 @@ export const ReorderRow = ({ id, updateOrder, children, ...props }: ReorderRowPr
...
-      {...dropProps}
-      ref={ref}
-      colSpan={2}
-      className={`dnd-row${isDragging ? ' dragging' : ''}${isDropTarget ? ' target' : ''}`}
-      {...props}
-    >
-      {children}
-      <td className='dnd-row-handle'>
-        <IvyIcon icon={IvyIcons.ChangeType} />
-      </td>
```
ivy-lli, line 55: I don't understand why this now needs to be inside of the cell... this Has nothing to do with the cell, you reorder the row not the cell!

packages/editor/src/components/parts/rest/rest-request/rest-target/RestProperties.tsx
```diff
@@ -10,16 +9,14 @@ export const RestProperties = () => {
   const knownProperties = useMeta('meta/rest/properties', undefined, []).data;
 
   return (
-    <PathCollapsible
-      label='Properties'
-      path='properties'
-      defaultOpen={!deepEqual(config.target.properties, defaultConfig.target.properties)}
-    >
+    <PathContext path='properties'>
```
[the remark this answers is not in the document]
ivy-edp: Because the PropertyTable is also used for the Accept-Properties and there was no Path set. There was only one path set called "headers".
grafik

packages/editor/src/components/parts/common/path/validation/ValidationRow.tsx
```diff
@@ -45,13 +45,17 @@ export const SelectableValidationRow = <TData extends object>({
   );
 };
 
-type ValidationReorderRowProps = ValidationRowProps & ReorderRowProps;
+type ValidationReorderRowProps<TData> = ValidationRowProps & ReorderRowProps & SelectRowProps<TData> & { colSpan: number };
```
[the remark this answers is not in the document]
ivy-edp: I need the colSpan for the SelectableReorderRow, where (I think) I pass it within the {...props}. The SelectableReorderRow needs the colSpan to ensure, that the Validation-Message is displayed correctly.
grafik

integrations/standalone/tests/pageobjects/Section.ts
```diff
@@ -33,6 +33,7 @@ export class Section extends Composite {
   }
 
   private static toggleButtonLocatorInternal(page: Page, label: string) {
-    return page.locator(`.collapsible-trigger:has-text("${label}")`);
+    const regexLabel = new RegExp(`^${label}$`);
+    return page.locator(`.collapsible-trigger:has-text("${label}")`, { hasText: regexLabel });
```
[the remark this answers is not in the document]
ivy-edp: Ah yes you are right, I will do it like this:
```
 return page.locator(`.collapsible-trigger`, { hasText: regexLabel });
```
I had to use regex because I added a header called "Accept-Properties" within the REST-Activity and with just (`.collapsible-trigger:has-text("Properties")` there where two results "Properties" and "Accept-Properties.

packages/editor/src/components/widgets/table/row/ReorderRow.tsx
ivy-edp: Nevermind maybe I could do it with an ReorderIcon-Wrapper as you suggested. 
grafik
grafik

What do you prefer?

packages/editor/src/components/parts/common/path/validation/ValidationRow.tsx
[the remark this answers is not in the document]
ivy-lli: Yes but I think then the SelectRowProps should already contain this colSpan prop.

packages/editor/src/components/parts/rest/rest-request/rest-target/RestProperties.tsx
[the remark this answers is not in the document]
ivy-lli: Ah didn't saw that. 👍

packages/editor/src/components/widgets/table/row/ReorderRow.tsx
ivy-lli: Both ways much better! I think I would go with the second one. But I woudn't call this just a ReorderIcon. Because if you do so, someone else could think he can just use this everywhere. Better something like ReorderWrapper or ReorderWrapperIcon...

packages/editor/src/components/parts/condition/ConditionTable.tsx
```diff
@@ -69,30 +76,56 @@ const ConditionTable = ({ data, onChange }: { data: Condition[]; onChange: (chan
...
+
+  const tableActions = showDeleteAction
+    ? [
+        {
+          label: 'Remove row',
+          icon: IvyIcons.Trash,
+          action: () => removeRow(table.getRowModel().rowsById[Object.keys(rowSelection)[0]].index)
+        }
+      ]
+    : [];
```
ivy-lli, line 90: not quite sure, the old impl only showed a delete button for rows that are unknown. if this logic here does this, ignore this comment :)

packages/editor/src/components/parts/condition/ConditionTable.tsx
```diff
@@ -69,30 +76,56 @@ const ConditionTable = ({ data, onChange }: { data: Condition[]; onChange: (chan
...
-          <ValidationReorderRow colSpan={2} key={row.id} id={row.original.fid} updateOrder={updateOrder} rowPathSuffix={row.index}>
-            {row.getVisibleCells().map(cell => (
-              <TableCell key={cell.id}>{flexRender(cell.column.columnDef.cell, cell.getContext())}</TableCell>
-            ))}
-          </ValidationReorderRow>
-        ))}
-      </tbody>
-    </Table>
+    <div>
+      {showDeleteAction && <Fieldset label=' ' controls={tableActions} />}
```
ivy-lli, line 94: Would be fine for me if this fieldset here has also 'Conditions' as label

packages/editor/src/components/parts/mail/MailAttachmentPart.tsx
```diff
@@ -29,70 +30,77 @@ const MailAttachmentsPart = () => {
...
   });
 
+  const tableActions =
+    table.getSelectedRowModel().rows.length > 0
+      ? [
+          {
+            label: 'Remove row',
+            icon: IvyIcons.Trash,
+            action: () => removeRow(table.getRowModel().rowsById[Object.keys(rowSelection)[0]].index)
+          }
```
ivy-lli, line 70: maybe instead of the removeRow function which is provided by the useResizableEditableTable we should just return this Remove ControlField object...

packages/editor/src/components/parts/mail/MailAttachmentPart.tsx
```diff
@@ -29,70 +30,77 @@ const MailAttachmentsPart = () => {
...
-                <TableCell key={cell.id}>{flexRender(cell.column.columnDef.cell, cell.getContext())}</TableCell>
-              ))}
-              <ActionCell actions={[{ label: 'Remove row', icon: IvyIcons.Trash, action: () => removeTableRow(row.index) }]} />
-            </ValidationRow>
-          ))}
-        </tbody>
-      </Table>
-      <TableAddRow addRow={addRow} />
+      <div>
+        {table.getSelectedRowModel().rows.length > 0 && <Fieldset label=' ' controls={tableActions} />}
```
ivy-lli, line 77: same here, I think it would be ok to have here a label 'Attachmets' or whatever for this fieldset

packages/editor/src/components/parts/mail/MailAttachmentPart.tsx
```diff
@@ -29,70 +30,77 @@ const MailAttachmentsPart = () => {
...
+                  <TableHeader key={header.id} colSpan={header.colSpan}>
+                    {flexRender(header.column.columnDef.header, header.getContext())}
+                  </TableHeader>
+                ))}
+              </ResizableHeader>
+            ))}
+          </thead>
+          <tbody>
+            {table.getRowModel().rows.map(row => (
+              <SelectableValidationRow row={row as Row<object>} colSpan={1} key={row.id} rowPathSuffix={row.index}>
```
ivy-lli, line 92: why is this 'as' needed?

packages/editor/src/components/parts/query/database/TableReadFields.tsx
```diff
@@ -78,9 +80,16 @@ export const TableReadFields = () => {
...
             {table.getHeaderGroups().map(headerGroup => (
-              <tr key={headerGroup.id}>
+              <tr
+                key={headerGroup.id}
+                onDoubleClick={() => {
+                  headerGroup.headers.forEach(header => {
+                    header.column.resetSize();
+                  });
+                }}
+              >
```
ivy-lli, line 90: don't have we a component for this?

packages/editor/src/components/parts/query/database/TableSort.tsx
```diff
@@ -57,23 +60,46 @@ export const TableSort = () => {
     () => [
       {
         accessorKey: 'name',
-        header: header => <SortableHeader header={header} name='Column' />,
+        header: header => <SortableHeader header={header} name='Column' seperator={true} />,
```
ivy-lli, line 63: I think this should also not be sortable as it is reorderable.

packages/editor/src/components/widgets/table/row/ReorderRow.tsx
```diff
@@ -37,22 +48,40 @@ export const ReorderRow = ({ id, updateOrder, children, ...props }: ReorderRowPr
...
+      >
+        {children}
+      </SelectRow>
+      <MessageRow colSpan={colSpan ? colSpan : 2} message={message} />
+    </>
   );
 };
+
+export const ReorderWrapperIcon = ({ children }: React.HTMLAttributes<HTMLTableCellElement>) => (
+  <div className='reorder-select-cell'>
```
ivy-lli, line 81: I think i would not call this 'cell' as it is not really a cell, I think I would call it 'reorder-select-icon' or 'reorder-select-handle'

ivy-lli (approved):
Yea! Thanks for the improvements 🥇 
Some questions and small suggestions:

packages/editor/src/components/parts/mail/MailAttachmentPart.tsx
ivy-edp: Yes, I can do that but do you maybe know a better name for the Fieldset than "Attachments". Because I think if we set a Fieldset to Attachments the word "Attachments" is appearing too much:
grafik

ivy-lli: maybe change the table header? or maybe this header is not even needed?

packages/editor/src/components/parts/condition/ConditionTable.tsx
ivy-edp: Yes with the "showDeleteAction" const this logic enables the same behavior as before 👍

packages/editor/src/components/parts/mail/MailAttachmentPart.tsx
ivy-edp: Good idea, I have now simply removed the header.
grafik

packages/editor/src/components/parts/mail/MailAttachmentPart.tsx
ivy-edp: Yes, however, I think I need to implement a function that returns this object instead of directly providing an object. This is because the removeRow function within the ControlField object requires an index that can only be passed from the component. So, if I provide a function where you can pass the index property and then return the corresponding ControlField, it should do the trick, correct?

ivy-edp: Nevermind, I made a thinking error. I can simply return the object without any issues 👍

packages/editor/src/components/parts/common/table/useResizableEditableTable.ts
```diff
@@ -83,7 +84,13 @@ const useResizableEditableTable = <TData>({ data, columns, onChange, emptyDataOb
     onChange(newData);
   };
 
-  return { table, rowSelection, setRowSelection, addRow, removeRow, showAddButton };
+  const removeRowAction: FieldsetControl = {
+    label: 'Remove row',
+    icon: IvyIcons.Trash,
+    action: () => removeRow(table.getRowModel().rowsById[Object.keys(rowSelection)[0]].index)
+  };
```
ivy-lli, line 91: Nice 👍

ivy-lli (approved):

PR #498, opened 2024-02-06
Add table column browser

ivy-edp:
I've added the Table Column Browser to the new Inscription View and removed the SQL Operations Browser. 

I haven't added integration tests yet, as that would require additional time to incorporate database tabs into the mock.html file. Do you think it's necessary to adjust the mock.html for testing the table column browser?

grafik


packages/editor/src/components/browser/tableCol/TableColBrowser.tsx
```diff
@@ -1,21 +1,136 @@
...
+  const columnMetas = useMeta('meta/database/columns', { context, database: config.query.dbName, table: config.query.sql.table }, []).data;
+
+  const [data, setData] = useState<Column[]>([]);
+
+  const [showHelper, setShowHelper] = useState(false);
+
+  useEffect(() => {
+    const select = config.query.sql.select;
+    console.log(columnMetas);
+    console.log(select);
```
ivy-lli, line 43: remove

packages/editor/src/components/browser/tableCol/TableColBrowser.tsx
```diff
@@ -1,21 +1,136 @@
...
+
+    const columnData = columnMetas.map<Column>(c => {
+      return { ...c, selected: select.includes(c.name) };
+    });
+    console.log(columnData);
```
ivy-lli, line 48: remove

packages/editor/src/components/browser/tableCol/TableColBrowser.tsx
```diff
@@ -1,21 +1,136 @@
...
+    setData(columnData.filter(c => c.selected === true));
```
ivy-lli, line 49: why split this map and filter? do in one step:
```ts
columnMetas.filter(c => c.select.includes(c.name));
```

or why is this selected needed?

ivy-lli: does this works too if you select all rows via the checkbox?

ivy-lli:
for the mock test: I don't know if the mock impl already returns data for this meta

packages/editor/src/components/browser/tableCol/TableColBrowser.tsx
ivy-edp: Well, I've written the code more compactly now, but I have to map it first and then filter it because I don't receive the 'select' information directly from the metas but instead through the useQueryData hook.

And you're right, that didn't work with the checkbox. I've adjusted it now so that it works with the checkbox as well.

ivy-lli (changes requested):
- don't change the table read fields logic. just simple don't filter if a '*' is in the array
- do not first map and then filter just only run a filter

packages/editor/src/components/browser/tableCol/TableColBrowser.tsx
```diff
@@ -1,21 +1,129 @@
...
+
+  const columnMetas = useMeta('meta/database/columns', { context, database: config.query.dbName, table: config.query.sql.table }, []).data;
+
+  const [data, setData] = useState<DatabaseColumn[]>([]);
+
+  const [showHelper, setShowHelper] = useState(false);
+
+  useEffect(() => {
+    const select = config.query.sql.select;
+    setData(select.length === 1 && select.indexOf('*') === 0 ? columnMetas : columnMetas.filter(c => select.includes(c.name)));
```
ivy-lli, line 42: instead of `? :` I would write this with if else, IMO much easier to read, and instead of index just compere the first element:

```ts
let columns = columnMetas;
if (select.length > 1 || select[0] !== '*') {
  columns = columnMetas.filter(c => select.includes(c.name)));
}
setData(columns);
```

ivy-lli (approved):
Just a small improvement left:

ivy-lli (approved):

PR #502, opened 2024-02-13
Allow to connect to an unsecure websocket

ivy-cst:
Even if the inscription client is served secure with https


ivy-cst:
Needed that I could debug my local engine which is not secure but using the build inscritiption clinet directly form https://jenkins.ivyteam.io/ to investigate this problem: https://github.com/axonivy/core/pull/4983

ivy-cst:
@ivy-lli

ivy-lmu (approved):

PR #504, opened 2024-02-16
XIVY-13091 Cleanup Browser Mock-Tests

ivy-edp:
i have divided the browser mock tests into 7 separate tests (one test file per browser) in order not to lose the overview


ivy-lli (approved):
I didn't have a look at the different test, as I assume they are just movements from the combined test file.
But looks good thanks 👍

PR #509, opened 2024-02-21
XIVY-11890: Add known Custom Fields to Table

ivy-edp:
I added Combocells to the Custom Fields Table, inwhich known Custom Fields can be selected. If a known Custom Field is already in the Table it is no longer displayed in the dropdown. Also after selecting a Custom Field via the dropdown the type is also automatically adjusted.
There are UT I had to skip, because the test didn't work anymore with the combobox. Will fix it in a later PR.
beispiel_customfields

### DIFF
--- a/packages/editor/src/components/parts/common/customfield/CustomFieldTable.tsx
+++ b/packages/editor/src/components/parts/common/customfield/CustomFieldTable.tsx
@@ -1,4 +1,4 @@
-import type { WfCustomField, CustomFieldConfigType } from '@axonivy/inscription-protocol';
+import type { WfCustomField, WorkflowType } from '@axonivy/inscription-protocol';
 import { CUSTOM_FIELD_TYPE } from '@axonivy/inscription-protocol';
 import { IvyIcons } from '@axonivy/editor-icons';
 import type { ColumnDef } from '@tanstack/react-table';
@@ -6,7 +6,6 @@ import { flexRender } from '@tanstack/react-table';
 import { memo, useMemo } from 'react';
 import type { SelectItem } from '../../../../components/widgets';
 import {
-  EditableCell,
   SelectCell,
   Table,
   TableHeader,
@@ -14,17 +13,19 @@ import {
   TableAddRow,
   SortableHeader,
   ScriptCell,
-  ResizableHeader
+  ResizableHeader,
+  ComboCell
 } from '../../../../components/widgets';
-import { useAction } from '../../../../context';
+import { useAction, useEditorContext, useMeta } from '../../../../context';
 import { SelectableValidationRow } from '../path/validation/ValidationRow';
 import { PathCollapsible } from '../path/PathCollapsible';
 import { useResizableEditableTable } from '../table/useResizableEditableTable';
+import { deepEqual } from '../../../../utils/equals';
 
 type CustomFieldTableProps = {
   data: WfCustomField[];
   onChange: (change: WfCustomField[]) => void;
-  type: CustomFieldConfigType;
+  type: WorkflowType;
 };
 
 const EMPTY_WFCUSTOMFIELD: WfCustomField = { name: '', type: 'STRING', value: '' } as const;
@@ -32,12 +33,21 @@ const EMPTY_WFCUSTOMFIELD: WfCustomField = { name: '', type: 'STRING', value: ''
 const CustomFieldTable = ({ data, onChange, type }: CustomFieldTableProps) => {
   const items = useMemo<SelectItem[]>(() => Object.entries(CUSTOM_FIELD_TYPE).map(([value, label]) => ({ label, value })), []);
 
+  const { context } = useEditorContext();
+
+  const predefinedCustomField: WfCustomField[] = useMeta('meta/workflow/customFields', { context, type: type }, []).data;
+
   const columns = useMemo<ColumnDef<WfCustomField>[]>(
     () => [
       {
         accessorKey: 'name',
         header: header => <SortableHeader header={header} name='Name' seperator={true} />,
-        cell: cell => <EditableCell cell={cell} />
+        cell: cell => (
+          <ComboCell
+            items={predefinedCustomField.filter(pcf => !data.find(d => d.name === pcf.name)).map(pcf => ({ value: pcf.name }))}
+            cell={cell}
+          />
+        )
       },
       {
         accessorKey: 'type',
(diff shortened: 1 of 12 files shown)


packages/editor/src/components/parts/common/customfield/CustomFieldTable.tsx
```diff
@@ -32,12 +34,43 @@ const EMPTY_WFCUSTOMFIELD: WfCustomField = { name: '', type: 'STRING', value: ''
...
+      return 'TEXT';
+    }
+    if (typeAsString === 'NUMBER') {
+      return 'NUMBER';
+    }
+    if (typeAsString === 'TIMESTAMP') {
+      return 'TIMESTAMP';
+    }
+    return 'STRING';
+  };
```
ivy-lli, line 51: looks like a useless conversion to me... Why is this needed?

packages/editor/src/components/parts/common/customfield/CustomFieldTable.tsx
```diff
@@ -32,12 +34,43 @@ const EMPTY_WFCUSTOMFIELD: WfCustomField = { name: '', type: 'STRING', value: ''
...
+
+  const predefinedCustomField: WfCustomField[] = useMeta(
+    'meta/workflow/customFields',
+    { context, type: toWorkflowType(path) },
```
ivy-lli, line 55: why? I think the path never be only 'Text' or 'Number' or 'Timestamp', so it always results in String...

packages/editor/src/components/parts/common/customfield/CustomFieldTable.tsx
```diff
@@ -32,12 +34,43 @@ const EMPTY_WFCUSTOMFIELD: WfCustomField = { name: '', type: 'STRING', value: ''
...
+    []
+  ).data.map<WfCustomField>(pcf => ({
+    name: pcf.name,
+    type: toWfFieldType(pcf.type),
```
ivy-lli, line 59: what does pcf mean?
why do we have to convert this type?

packages/editor/src/components/parts/common/customfield/StartCustomFieldTable.tsx
```diff
@@ -17,20 +28,30 @@ type StartCustomFieldTableProps = {
 const EMPTY_STARTCUSTOMSTARTFIELD: StartCustomStartField = { name: '', value: '' } as const;
 
 const StartCustomFieldTable = ({ data, onChange }: StartCustomFieldTableProps) => {
+  const { context } = useEditorContext();
+  const path = usePath();
+  const predefinedCustomField: ComboboxItem[] = useMeta(
+    'meta/workflow/customFields',
+    { context, type: toWorkflowType(path) },
```
ivy-lli, line 35: it is always type: 'START' here right?

packages/editor/src/components/parts/common/table/useResizableEditableTable.ts
```diff
@@ -31,20 +38,24 @@ const useResizableEditableTable = <TData>({ data, columns, onChange, emptyDataOb
...
+                [columnId]: value
+              };
+            }
+            return row;
+          });
+          if (!deepEqual(updatedData[updatedData.length - 1], emptyDataObject) && rowIndex === data.length - 1) {
+            onChange([...updatedData, emptyDataObject]);
+          } else {
+            onChange(updatedData);
+          }
```
ivy-lli, line 58: instead of blow this up, I would create a updateData function outside of this meta object and override it with the specialUpdateData stuff if available... Or maybe even more can be shared between those two?

packages/protocol/src/data/inscription.ts
```diff
@@ -209,6 +206,11 @@ export interface ContentObject {
 export interface MapStringString {
   [k: string]: string;
 }
+export interface CustomFieldMeta {
+  description: string;
+  name: string;
+  type: string;
```
ivy-lli, line 212: this should maybe more typed?

ivy-lli: I think we know what different types here exist right?

packages/editor/src/components/parts/common/customfield/CustomFieldTable.tsx
ivy-lli: and we already have the type of the customfield table in the props right?

packages/editor/src/components/parts/common/customfield/CustomFieldTable.tsx
ivy-edp: Yes, I have now removed that conversion. The version you reviewed still has an outdated inscription.ts. Now I receive a WfCustomField array directly from "meta/workflow/customFields". Accordingly, I no longer need this. Sorry, I meant to add this as a reviewer for this PR just now.

packages/editor/src/components/parts/common/customfield/CustomFieldTable.tsx
ivy-edp: I am not sure if I understand correctly, but in this path, it's not about Text, Number, or Timestamp, but rather 'START, CASE, TASK'. This is necessary because there are different predefined custom fields depending on whether you're on a start, case, or task.

packages/editor/src/components/parts/common/customfield/CustomFieldTable.tsx
ivy-edp: No longer needed. I removed this mapping 👍

packages/editor/src/components/parts/common/customfield/StartCustomFieldTable.tsx
ivy-edp: Yes, you're right, the StartCustomFieldTable is effectively only used in the RequestPart. Should I just always set START as the type in this case?

packages/protocol/src/data/inscription.ts
ivy-edp: Does not exist anymore 👍. Instead of CustomFieldMeta, there is now customFieldSchema, which is of type "WfCustomField". If you're interested, this is the PR for the core changes.
https://github.com/axonivy/core/pull/5006

packages/editor/src/components/parts/common/table/useResizableEditableTable.ts
ivy-edp: do you mean something like this?
grafik

ivy-lli: Yes I had something like this in mind

packages/editor/src/components/parts/common/customfield/CustomFieldTable.tsx
```diff
@@ -1,43 +1,55 @@
...
 
 const EMPTY_WFCUSTOMFIELD: WfCustomField = { name: '', type: 'STRING', value: '' } as const;
 
 const CustomFieldTable = ({ data, onChange, type }: CustomFieldTableProps) => {
   const items = useMemo<SelectItem[]>(() => Object.entries(CUSTOM_FIELD_TYPE).map(([value, label]) => ({ label, value })), []);
 
+  const { context } = useEditorContext();
+  const path = usePath();
+
+  const predefinedCustomField: WfCustomField[] = useMeta('meta/workflow/customFields', { context, type: toWorkflowType(path) }, []).data;
```
ivy-lli, line 40: toWorkflowType IMO not needed, type is already there from the component props

packages/editor/src/components/parts/common/customfield/StartCustomFieldTable.tsx
ivy-lli: yes

packages/editor/src/components/parts/common/info/Information.tsx
ivy-lli, line 1: changes not needed, as you don't need to export the `toWorkflowType` I think

ivy-lli (commented):
Much better, but I still think you can simply use the prop type and don't need this toWorkflowType method...

packages/editor/src/components/parts/common/customfield/CustomFieldTable.tsx
ivy-edp: Ah now I get it, thanks. I will remove the toWorkflowType method from CustomFieldTable 👍

ivy-lli (approved):

PR #510, opened 2024-02-21
Load and initialize Monaco editor lazily

martin-fleck-at:
- Split monaco-related classes into dedicated chunk
- Lazy load files and modules through 'await import'
- Add placeholder text while loading Monaco editor in React

---

On the master (first commit adds time measurements) most startup times were around 160 - 250ms:
```
MonacoUtil.initStandalone: 31.322998046875 ms
monaco-editor-util.ts:77 MonacoEditorUtil.initMonaco: 2.029052734375 ms
ivy-script-client.ts:10 startWebSocketClient: 70.7958984375 ms
ivy-script-client.ts:23 IvyScriptLanguage.startClient: 5.37109375 ms
index.tsx:46 *** start: 217.096923828125 ms
```

With this commit (second commit) most startup times are around 50 - 90ms. However, as you can see, the overall starting time was increased as well. I assume that is because of how the files are loaded asynchronously. If necessary, I can investigate this further:

```
IvyScriptLanguage.startWebSocketClient: 33.200927734375 ms - Waiting for Monaco Editor...
index.tsx:57 *** start: 61.468017578125 ms
monaco-util.ts:21 MonacoUtil.initStandalone: 22.028076171875 ms
monaco-editor-util.ts:121 configureMonaco: 1.72998046875 ms
index.tsx:23 initMonaco: 350.2490234375 ms
ivy-script-client.ts:10 IvyScriptLanguage.startWebSocketClient: 350.27099609375 ms
ivy-script-client.ts:24 IvyScriptLanguage.startClient: 265.011962890625 ms
```
Please note that I remove the time measurements again in the third commit so if you want to try this out, I suggest to use the second commit.

While loading the editor we display a simple message but any React component could be used:
lazy-editor
The loading is usually too fast for you to expand to an editor, so I suggest to add a delay, e.g., in `configureMonaco`:
```
const def = new Deferred<void>();
setTimeout(() => def.resolve(), 5000);
await def.promise;
```


martin-fleck-at:
I converted this to a draft for now to have a look at the tests.

martin-fleck-at:
Anybody got an idea why the tests are failing? Locally they run through:
image

ivy-lli:
Hi @martin-fleck-at
Seems to be flaky tests. But I saw that they open the url twice so, if it takes a long time (maybe because the build node is on high pressure) then it runs in a timeout of 30s.
I will shorty merge the PR with a fix for the open url twice: https://github.com/axonivy/inscription-client/pull/512

packages/editor/package.json
```diff
@@ -15,6 +15,7 @@
   ],
   "dependencies": {
     "@axonivy/editor-icons": "11.3.0-next",
+    "@axonivy/inscription-core": "11.3.0-next",
```
ivy-lli, line 18: why is this dependency needed, if the only thing that is now needed from this package is the Deferred util, and it is currently only needed in the editor? Why not move this Deferred util directly to the editor package?
Why is this Deferret util even needed? I think I don't fully understand the purpose of it 🙈

martin-fleck-at: We can absolutely move it into the `editor` package as well if you prefer. It really is just such a fundamental utility class that we wanted to put it somewhere more general but unless somebody else uses it we can definitely move it into the editor package.

The idea behind the `Deferred` is a common pattern in many frameworks but it is basically just convenience "manager" for a single promise. When you create a promise you get the `resolve` and `reject` functions that you can call to work with the promise but those functions are not exposed and cannot be called from the outside so you would need to implement all your logic within the promise constructor. So just exposing those functions already gives you a lot more flexibility with the code as you decouple the promise creation from its resolution. You can put additional meta information in the Deferred as well such as the current state of the promise.

So if you look at our use case below, you'll see we immediately create the Deferred object so anybody can already call `getInstance()` to get the future promise without any async code already working on resolving that promise. That only happens after `configureMonaco` is called and we actually resolve the promise. That is not to say you cannot do this differently.

ivy-lli: I think I would move it into the editor package for now 👍

ivy-lli: Thanks for your explanation 👍

martin-fleck-at: I moved it into the editor package in a new commit.

ivy-lli (approved):

PR #514, opened 2024-02-22
XIVY-13253: Prevents display of error on new line in parameter table

ivy-edp:
To prevent errors when adding a new row to the parameter table, a default value is now added to the name. 

I've named it "change_this_name" to signal that the name needs to be changed. If the name isn't changed and the component is reloaded or the selection is lost, the new row will be removed from the table. I haven't named it "parameter" because I can imagine that this name might often be forgotten to be updated. 
What do you think?


ivy-lli (approved):
When is it removed again?

ivy-edp:
It will remove it if the row has set Name: "change_this_name", Type: "String", Description: ""
Every Row that looks like this, will be removed, after reloading the component or when the selection is reset by pressing on the header:
beispiel_add_parameter(2)

ivy-lli:
ah the logic is the empty object. nice 👍

PR #518, opened 2024-02-27
Fix combocell

ivy-edp:
I've adjusted the combobox for use with the combocell, as the new table behavior prevents direct adoption. To avoid creating a new combobox in the combocell (since most of it can be reused), I added the "asCell" flag. Hope that's alright. 

Also, I've re-added the tests for the customfieldstable and removed the "change_this_name" as it is no longer necessary if the changes from the core are merged.

### DIFF
--- a/packages/editor/src/components/widgets/combobox/Combobox.tsx
+++ b/packages/editor/src/components/widgets/combobox/Combobox.tsx
@@ -25,6 +25,7 @@ export type ComboboxProps<T extends ComboboxItem> = Omit<ComponentProps<'input'>
   onChange: (change: string) => void;
   macro?: boolean;
   browserTypes?: BrowserType[];
+  updateOnInputChange?: boolean;
 };
 
 const Combobox = <T extends ComboboxItem>({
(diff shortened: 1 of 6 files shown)


packages/editor/src/components/widgets/combobox/Combobox.tsx
```diff
@@ -35,6 +36,7 @@ const Combobox = <T extends ComboboxItem>({
   onChange,
   macro,
   browserTypes,
+  asCell,
```
ivy-lli, line 39: I think I understand why it is needed. It is because the cell and the combobox currently both "save" the value if the focus is removed from it and the new value need to go through both of them.

What I don't like is this parameter name. If I see this combobox I would not understand it's purpose until I read the code. Nether would I understand what this parameter changes if I would use this combobox somewhere. A parameter should be named after the purpose what it's doing. In this case it defines when the onChange is triggered, directly after each change or after a blur/focus loose. So it should be called something like 'directUpdate' or 'directOnChange'... Maybe you have a better name in mind.

ivy-edp: Ah yes, thats a good point. I think I will rename the parameter to "updateOnInputChange" 👍

ivy-lli (approved):

PR #520, opened 2024-02-29
XIVY-12913: Add help-button to Editor-Header

ivy-edp:
As soon as Anais has created the symbol, I will change it to this one:
grafik


packages/editor/src/components/editors/InscriptionEditor.tsx
```diff
@@ -42,6 +43,16 @@ const Header = (props: EditorProps) => {
             {data.name.length > 0 && ` - ${data.name}`}
           </div>
         </div>
+
+        {editorContext.type.helpUrl !== undefined && (
```
ivy-lli, line 47: what is, when the helpUrl is empty ("")?
I think I would not show a help button if it is like this...

ivy-lli (approved):
Looks good, one small improvement:

packages/editor/src/components/editors/InscriptionEditor.tsx
```diff
@@ -42,6 +44,16 @@ const Header = (props: EditorProps) => {
             {data.name.length > 0 && ` - ${data.name}`}
           </div>
         </div>
+
+        {(helpUrl !== undefined || helpUrl !== '') && (
```
ivy-lli, line 48: why '||' ?
if now the helpUrl is '' the or is true so it will render the help icon

PR #521, opened 2024-03-01
XIVY-11709: First version of thirdparty-editors

ivy-edp:
I have now added the three third-party editors (Rule, Intermediate-ThridPartyWait, Start-ThirdPartyStart). I have removed the Java class selector from both the Start and Event tabs for thirdparty-editors.

@ivy-lli  I have created a new folder "third-party" where I have placed all the thirdparty-editors. There are actually only three editors, which I could also simply incorporate into the "other-editors" TSX file. What do you think?

beispiel_thirdParty

### DIFF
--- a/packages/editor/src/components/editors/third-party/interface.tsx
+++ b/packages/editor/src/components/editors/third-party/interface.tsx
@@ -1,0 +1,18 @@
+/* eslint-disable react/jsx-key */
+import { IvyIcons } from '@axonivy/ui-icons';
+import type { ElementType } from '@axonivy/inscription-protocol';
+import type { ReactNode } from 'react';
+import { memo } from 'react';
+import InscriptionEditor from '../InscriptionEditor';
+import { useGeneralPart, useProgramInterfaceStartPart, useConfigurationPart } from '../../../components/parts';
+
+const ThirdPartyProgramInterfaceEditor = memo(() => {
+  const name = useGeneralPart();
+  const start = useProgramInterfaceStartPart({ thirdParty: true });
+  const configuration = useConfigurationPart();
+  return <InscriptionEditor icon={IvyIcons.Extension} parts={[name, start, configuration]} />;
+});
+
+export const thirdPartyInterfaceActivityEditors = new Map<ElementType, ReactNode>([
+  ['ThirdPartyProgramInterface', <ThirdPartyProgramInterfaceEditor />]
+]);
--- a/packages/editor/src/components/editors/third-party/intermediate.tsx
+++ b/packages/editor/src/components/editors/third-party/intermediate.tsx
@@ -1,0 +1,18 @@
+/* eslint-disable react/jsx-key */
+import { IvyIcons } from '@axonivy/ui-icons';
+import type { ElementType } from '@axonivy/inscription-protocol';
+import type { ReactNode } from 'react';
+import { memo } from 'react';
+import InscriptionEditor from '../InscriptionEditor';
+import { useConfigurationPart, useEventPart, useGeneralPart, useOutputPart, useTaskPart } from '../../../components/parts';
+
+const ThirdPartyWaitEventEditor = memo(() => {
+  const name = useGeneralPart();
+  const event = useEventPart({ thirdParty: true });
+  const configuration = useConfigurationPart();
+  const task = useTaskPart({ type: 'wait' });
+  const output = useOutputPart();
+  return <InscriptionEditor icon={IvyIcons.Extension} parts={[name, event, configuration, task, output]} />;
+});
+
+export const thirdPartyIntermediateEventEditors = new Map<ElementType, ReactNode>([['ThirdPartyWaitEvent', <ThirdPartyWaitEventEditor />]]);
--- a/packages/editor/src/components/editors/third-party/start.tsx
+++ b/packages/editor/src/components/editors/third-party/start.tsx
@@ -1,0 +1,17 @@
+/* eslint-disable react/jsx-key */
+import { IvyIcons } from '@axonivy/ui-icons';
+import type { ElementType } from '@axonivy/inscription-protocol';
+import type { ReactNode } from 'react';
+import { memo } from 'react';
+import InscriptionEditor from '../InscriptionEditor';
+
+import { useGeneralPart, useProgramStartPart, useConfigurationPart } from '../../../components/parts';
+
+const ThirdPartyProgramStartEditor = memo(() => {
+  const name = useGeneralPart();
+  const start = useProgramStartPart({ thirdParty: true });
+  const configuration = useConfigurationPart();
+  return <InscriptionEditor icon={IvyIcons.Extension} parts={[name, start, configuration]} />;
+});
+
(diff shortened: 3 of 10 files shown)


packages/editor/src/components/parts/program/activity/ProgramInterfaceStartPart.tsx
```diff
@@ -32,7 +32,9 @@ const ProgramInterfaceStartPart = () => {
 
   return (
     <>
-      <JavaClassSelector javaClass={config.javaClass} onChange={change => update('javaClass', change)} type='ACTIVITY' />
+      {(thirdParty == undefined || thirdParty == false) && (
```
ivy-lli, line 35: first why only two '==' ? in js you should always use three...
And second, you can also use `thirdParty!!` instead

packages/editor/src/components/editors/third-party/start.tsx
```diff
@@ -0,0 +1,17 @@
+/* eslint-disable react/jsx-key */
+import { IvyIcons } from '@axonivy/ui-icons';
+import type { ElementType } from '@axonivy/inscription-protocol';
+import type { ReactNode } from 'react';
+import { memo } from 'react';
+import InscriptionEditor from '../InscriptionEditor';
+
+import { useGeneralPart, useProgramStartPart, useConfigurationPart } from '../../../components/parts';
+
+const ProgramStartEditor = memo(() => {
```
ivy-lli, line 10: I would call those here also ThirdPartyProgramStartEditor

packages/editor/src/components/editors/third-party/intermediate.tsx
```diff
@@ -0,0 +1,18 @@
+/* eslint-disable react/jsx-key */
+import { IvyIcons } from '@axonivy/ui-icons';
+import type { ElementType } from '@axonivy/inscription-protocol';
+import type { ReactNode } from 'react';
+import { memo } from 'react';
+import InscriptionEditor from '../InscriptionEditor';
+import { useConfigurationPart, useEventPart, useGeneralPart, useOutputPart, useTaskPart } from '../../../components/parts';
+
+const WaitEventEditor = memo(() => {
```
ivy-lli, line 9: ThirdPartyWaitEventEditor

packages/editor/src/components/editors/third-party/start.tsx
```diff
@@ -0,0 +1,17 @@
...
+  const name = useGeneralPart();
+  const start = useProgramStartPart({ thirdParty: true });
+  const configuration = useConfigurationPart();
+  return <InscriptionEditor icon={IvyIcons.StartProgram} parts={[name, start, configuration]} />;
```
ivy-lli, line 14: I would use the extension icon like we do in the process palette

packages/editor/src/components/editors/third-party/intermediate.tsx
```diff
@@ -0,0 +1,18 @@
...
+  const name = useGeneralPart();
+  const event = useEventPart({ thirdParty: true });
+  const configuration = useConfigurationPart();
+  const task = useTaskPart({ type: 'wait' });
+  const output = useOutputPart();
+  return <InscriptionEditor icon={IvyIcons.ClockOutline} parts={[name, event, configuration, task, output]} />;
```
ivy-lli, line 15: I would use the extension icon like we do in the process palette

packages/editor/src/components/editors/third-party/interface.tsx
```diff
@@ -0,0 +1,18 @@
+/* eslint-disable react/jsx-key */
+import { IvyIcons } from '@axonivy/ui-icons';
+import type { ElementType } from '@axonivy/inscription-protocol';
+import type { ReactNode } from 'react';
+import { memo } from 'react';
+import InscriptionEditor from '../InscriptionEditor';
+import { useGeneralPart, useProgramInterfaceStartPart, useConfigurationPart } from '../../../components/parts';
+
+const RuleInterfaceEditor = memo(() => {
```
ivy-lli, line 9: ThirdPartyProgramInterfaceEditor

there can also be other thirdparty program interface activities...

packages/editor/src/components/editors/third-party/interface.tsx
```diff
@@ -0,0 +1,18 @@
...
+  const name = useGeneralPart();
+  const start = useProgramInterfaceStartPart({ thirdParty: true });
+  const configuration = useConfigurationPart();
+  return <InscriptionEditor icon={IvyIcons.ProgramOutline} parts={[name, start, configuration]} />;
```
ivy-lli, line 13: I would use the extension icon like we do in the process palette

ivy-rew (approved):
from the process-domain perspective; it looks perfect .. just what I need.
code looks easy and well organized too ... but @ivy-lli should have the last word about this :)

ivy-lli (approved):
Good. 
For what the editors can and what not is @ivy-rew more the source of truth, as I'm not very familiar with this editors...

packages/editor/src/components/parts/program/activity/ProgramInterfaceStartPart.tsx
ivy-edp: thanks, but if I set thirdParty!! I get the following error:
grafik

ivy-lli: or `!!thirdParty`

ivy-lli: ah you check for falsy, yes of course then it is just `!thirdParty`

packages/editor/src/components/parts/program/activity/ProgramInterfaceStartPart.tsx
```diff
@@ -32,8 +32,8 @@ const ProgramInterfaceStartPart = ({ thirdParty }: { thirdParty?: boolean }) =>
 
   return (
     <>
-      {(thirdParty == undefined || thirdParty == false) && (
-        <JavaClassSelector javaClass={config.javaClass} onChange={change => update('javaClass', change)} type='ACTIVITY' />
+      {!thirdParty && (
+        <JavaClassSelector javaClass={config.javaClass} onChange={change => update('javaClass', change)} type='INTERMEDIATE' />
```
ivy-lli, line 36: are you sure about this type change?

ivy-edp: yes it works like that, I was also a bit confused but I checked it. 👍

ivy-rew: isn't this the type you will use the show already implemented beans in the project and to for the 'create' action?
ch.ivyteam.ivy.lsp.inscription.meta.program.impl.DefaultProgramMeta.types(ProgramInterfacesRequest)
... if so the type matters. If it doesn't match the 'element' nature, it will list and create invalid java impls.

ivy-edp: Oops, sorry, yes you are right. A copy past error. I have now changed it back to ACTIVITY 👍

PR #522, opened 2024-03-01
XIVY-13508: Add placeholder to monaco-editor

ivy-edp:
I have implemented the option to add a placeholder to the Monaco editor. Currently, I have structured it so that the placeholder text can be passed into the CodeEditor as a parameter. At the moment, the placeholder "Press CTRL + SPACE for auto-completion" is simply present in all CodeEditorCells. However, it might make sense to set this placeholder for all singleLineCodeEditors or even for all Monaco editors. What do you think?
beispiel_placeholder

### DIFF
--- a/packages/editor/src/components/widgets/code-editor/CodeEditor.css
+++ b/packages/editor/src/components/widgets/code-editor/CodeEditor.css
@@ -14,3 +14,16 @@
 .code-editor .code-input:focus-within {
   border: var(--activ-border);
 }
+
+.code-editor .monaco-placeholder {
+  position: absolute;
+  display: none;
+  white-space: pre-wrap;
+  top: 12px;
+  left: 11px;
+  color: var(--N500);
+  font-family: Consolas, 'Courier New', monospace;
+  pointer-events: none;
+  user-select: none;
+  z-index: 10;
+}
--- a/packages/editor/src/components/widgets/code-editor/CodeEditor.css
+++ b/packages/editor/src/components/widgets/code-editor/CodeEditor.css
@@ -25,5 +25,10 @@
   font-family: Consolas, 'Courier New', monospace;
   pointer-events: none;
   user-select: none;
-  z-index: 10;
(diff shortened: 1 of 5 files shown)


ivy-edp:
Unfortunately, there isn't a built-in placeholder option in the Monaco editor, so I had to invent this functionality myself using a reference and styling. Here are the two sources I found on this topic and used as references:
https://github.com/microsoft/monaco-editor/issues/568
https://stackblitz.com/edit/react-ts-6egvre?file=App.tsx

ivy-lli (approved):
Ahh crazy that this is not built in, but I think a fine solution 👍 
I think the placeholder should be always the same on each editor. So I think there is no need for a property.

ivy-edp:
Perfect, than I will remove all the placeholder-properties and write the Placeholder "Press CTRL + SPACE for auto-completion" directly into the placeholder-div inside the code editor. So now not only all SingleLineEditors have this place holder but all the other monaco-editors too.

packages/editor/src/components/widgets/code-editor/CodeEditor.css
```diff
@@ -25,5 +25,10 @@
   font-family: Consolas, 'Courier New', monospace;
   pointer-events: none;
   user-select: none;
-  z-index: 10;
+  z-index: 2;
+}
+
+.code-editor .with-lineNumbers {
+  left: 35px;
+  font-size: 12px;
```
ivy-lli, line 33: is the font-size different for full size monaco?

ivy-edp: Well not really. It is also 12px but for some reason the Placeholder-Text is way to big if I don't set the font-size.

with no font-size:
grafik
with font-size:
grafik

ivy-lli: Then I think you can set it to the default placeholder css class right? would make more sense to me.

PR #526, opened 2024-03-05
Adjust icons

ivy-edp:


packages/editor/src/components/editors/third-party/interface.tsx
```diff
@@ -10,7 +10,7 @@ const ThirdPartyProgramInterfaceEditor = memo(() => {
   const name = useGeneralPart();
   const start = useProgramInterfaceStartPart({ thirdParty: true });
   const configuration = useConfigurationPart();
-  return <InscriptionEditor icon={IvyIcons.Extension} parts={[name, start, configuration]} />;
+  return <InscriptionEditor icon={IvyIcons.Rule} parts={[name, start, configuration]} />;
```
ivy-lli, line 13: what about other thirdparty interface activities? they now all have the rule icon...
I would rather favor the extension icon which is true for all of them (maybe not ideal), instead of the rule icons which is wrong for some of them...

ivy-edp: I see your point. We considered using the same icon for the rule as in the process editor, but you are right, solving it like this, would misalign with other third-parties. 
@ivy-rew, would you be okay, when we keep the extions-icon for the rule? 

Otherwise one easy solution could involve checking within the InscriptionEditor if the shortLabel is 'Rule' and applying the corresponding icon. What are your thoughts?
grafik

ivy-lli: Very ugly if statement, and in my eyes not worth the tech depth... But yes when then something like that.
Is there no better solution for this?

ivy-edp: Yes, I agree 😅 I think the only "clean" way to solve this problem, is to introduce a new InscriptionType like "ThirdPartyProgramInterfaceRule" and than I could create a new const ThridPartyRuleInterfaceEditor, where I set the Icon to "Rule" 🤔 However, I am not sure if the effort required to overhaul everything for just an icon is worth the trouble.

ivy-lli: IMO there are only two ways to solve this:
1. improve support for third-party elements (let them also define an icon, and then this icon also needs to be defined via API and should also work in the process editor)
2. finally define what the rule element is. 
  - a third-party element (no longer include it per default within the engine), and remove all special if's for it
  - a normal element which is included per default with it's own type

ivy-edp: Yes, I agree with you. Perhaps that's a topic we can briefly touch upon tomorrow after the HTML editor discussion. Otherwise, for this pull request, I'll revert the icon for the third-party back to 'extension' and simply merge the help icon change.

ivy-lli: I think this an own story, because of the result of the decision more or less work needs to be done...

ivy-lli (approved):

PR #529, opened 2024-03-12
XIVY-13506: Fix overflow of header inside autocompletion-details

ivy-edp:
With this small fix, I can prevent the header text from overlapping other elements. Unfortunately, I don't really have the energy today to  figure out, how to send a shorter type for the header in the core. If you still wish to have the whole thing fixed in the core, I would create a follow-up story and address it after my vacation.


ivy-lli:
I will have a look at it tomorrow or on thursday 👍

ivy-lli:
As this type header can be big anyways, I think it makes sense to fix the style and not change the type in the lsp itself. 
I've fixed the header type height, so it should no longer overflow

PR #532, opened 2024-03-19
Improve lazy-loading of Monaco editor

martin-fleck-at:
- Use more dynamic imports for Monaco editor
-- Expose dynamically imported
-- Use type imports where possible to avoid full resolution
- Expose as much options for outside configuration as possible
- Avoid unnecessary boolean return type
- Avoid loading Monaco editor only for key codes
- Align react-query versions with process-editor-client


packages/core/src/ivy-script-client.ts
```diff
@@ -1,14 +1,14 @@
 import { createWebSocketConnection, type Connection } from '@axonivy/jsonrpc';
 
 export namespace IvyScriptLanguage {
-  export async function startWebSocketClient(url: string, isMonacoReady: Promise<boolean>): Promise<any> {
+  export async function startWebSocketClient(url: string, isMonacoReady: Promise<any>): Promise<any> {
```
ivy-lli, line 4: why not Promise<void>?

ivy-lli (commented):
Looks good. But maybe a quick explanation would be great.
More and more I think we need a separate package with all this monaco stuff in it:
- Setup monaco environment
- Setup lsp connection
- Provide react monaco component
- ...

packages/core/src/ivy-script-client.ts
martin-fleck-at: > why not Promise?

I'm not sure I fully understand but the return type that we typically use here is of `monacoEditorReact.Monaco` but since we are not actually awaiting the result in this function, I typed it any (instead of `void` or `boolean` previously) mainly to ease changes and avoid conversions like `MonacoEditorUtil.getInstance().then(() => {})` but also highlight the fact that we do not care about the return type.

ivy-lli: sorry not everything was posted... 
I wanted to write `Why not Promise<void>`?

martin-fleck-at:
> Looks good. But maybe a quick explanation would be great. More and more I think we need a separate package with all this monaco stuff in it:
> 
> * Setup monaco environment
> * Setup lsp connection
> * Provide react monaco component
> * ...

I agree that in the long run that might make things easier. However, some things are little bit more intertwined as they manipulate the global `MonacoEnvironment`, e.g., `@monaco-editor/react` provides a loader to setup the overall monaco and only then we can properly setup the workers in the environment. The good thing is, we get errors and warnings fast if something is not done correctly ;-) So do you think we should do this as part of this PR or could this be a follow-up? I know the setup is important but it would be great to see if the lazy loading actually brings that much benefit.

I'm happy to provide additional explanation: Documentation, code comments, etc.? I tried to separate out things into dedicated functions as much as possible and add comments here and there. But the basic steps are: initialize the monaco API, configure the language client ("VS Code API") connection with the monaco API, add worker creation to the `MonacoEnvironment`, and register our own languages and themes against the monaco API. With that, the `monaco` namespace is ready to be used and a React MonacoEditor can be created.

As for the lazy loading: The goal is to async all the heavy imports into promises so they are not loaded when the script is parsed initially but only once we actually trigger the loading function.

packages/core/src/ivy-script-client.ts
martin-fleck-at: Does my post above answer the question? Basically we want to avoid unnecessary conversions.

ivy-lli:
Yes, I think, I understood the changes.

Yes, can definitely be in a later PR.

ivy-lli (approved):

ivy-lli:
I will merge this now, ok? @martin-fleck-at

martin-fleck-at:
@ivy-lli Yes, thank you!

martin-fleck-at:
@ivy-lli Do you know how long it usually takes for the new version to be available so it can be consumed in the process editor?

PR #557, opened 2024-04-16
XIVY-13806: Rename, move or delete multiple tabs

ivy-edp:
Renamings: 
- User Dialog / User Task: Call → Dialog
- Program Start: Start →  Java Bean
- Trigger: Trigger → Process
- Call sub: Process call → Process
- Output → Output Data
- Progamm Activity:  Start → Java Bean
- Data Cache → Cache

Move: 
- Script: move script to output
- Error end: move script to error tab
- WS, Rest, db: data cache (or new just cache) move to the end (hint: cache was already at the end with WS and with Rest cache-tab was not there at all)
- Task Intermedate: output move to the end
    
Remove (as a result of the move):
- ScriptPart (was only used for Script)
- OutputScriptPart (was only used for Error end)

Open ToDo's (in next PR): 
- WS, Rest: output / code mapping to output tab (create error section)


packages/editor/src/components/editors/event/intermediate.tsx
```diff
@@ -16,11 +16,11 @@ import {
 
 const TaskSwitchEventEditor = memo(() => {
   const name = useGeneralPart();
-  const output = useOutputPart();
   const task = useTaskPart();
   const casePart = useCasePart();
   const endPage = useEndPagePart();
-  return <InscriptionEditor icon={IvyIcons.Task} parts={[name, output, task, casePart, endPage]} />;
+  const output = useOutputPart();
+  return <InscriptionEditor icon={IvyIcons.Task} parts={[name, task, casePart, endPage, output]} />;
```
ivy-lli, line 23: This was not defined right? That we move the output tab on this element?
Because I don't know if this is valid. @ivy-rew what do you think here?
This order has a meaning. It defines what is done on an element one after another: https://developer.axonivy.com/doc/11.3/designer-guide/process-modeling/process-elements/index.html

> The inscription view is divided into multiple sections and the order of the sections indicates the sequence of processing. For example in the figure below, the Output mapping is performed before the code in the third section is executed.

By the way when I see this screenshot and the text above. We should also check/update the doc

packages/editor/src/components/parts/output/OutputPart.tsx
```diff
@@ -8,22 +8,22 @@ import { MappingPart, PathCollapsible, ValidationFieldset } from '../common';
...
-    reset: { dirty, action: () => resetOutput(!options?.hideCode) },
-    content: <OutputPart showCode={!options?.hideCode} additionalBrowsers={options?.additionalBrowsers} />
+    reset: { dirty, action: () => resetOutput(options?.enableSudo) },
+    content: <OutputPart enableSudo={options?.enableSudo} additionalBrowsers={options?.additionalBrowsers} />
   };
 }
 
-const OutputPart = (props: { showCode?: boolean; additionalBrowsers?: BrowserType[] }) => {
-  const { config, defaultConfig, update } = useOutputData();
+const OutputPart = (props: { enableSudo?: boolean; additionalBrowsers?: BrowserType[] }) => {
```
ivy-lli, line 25: I would still call it 'showSudo'. Enable means more for me, that it is here per default but disabled...

ivy-lli (commented):
Looks good. Not sure of the TaskSwitchEvent change...

packages/editor/src/components/editors/event/intermediate.tsx
ivy-edp: Discussed at the desk with @ivy-lli & @ivy-rew: We move the output tab to the end so that it always appears at the end (for all elements).

ivy-lli (approved):

PR #561, opened 2024-04-17
XIVY-13806: In Db-Activity move error tab outside

ivy-edp:
I noticed that there is also an 'Error Section' in the query tab inside the database-activity. In order to maintain consistency with the other activities (WS & Rest), I suggest relocating the error section to its own tab as well. @ivy-lli & @ivy-rew What do you think?

grafik


ivy-rew:
> I noticed that there is also an 'Error Section' in the query tab inside the database-activity. In order to maintain consistency with the other activities (WS & Rest), I suggest relocating the error section to its own tab as well. @ivy-lli & @ivy-rew What do you think?
> 
> grafik

yes, I like that. The DB-query is anyway overloaded with features, so moving errors out and make the ui consistent sounds great.

ivy-lli (approved):
Especially as this error is always visible for all query kinds, I think it makes a lot of sense to streamline it with the other elements.

PR #562, opened 2024-04-17
Fix inscription screenshots

ivy-edp:
After the name change, some screenshots no longer worked. Here is a first fix. After getting feedback from bruno, I will finalize the documentation.


ivy-rew (approved):
thanks Elio

ivy-lli (approved):

PR #564, opened 2024-04-22
XIVY-13987: Fix error if trigger-checkbox is selected with no task

ivy-edp:
I have now customized the whole thing so that if no task has been selected yet, the same message appears as in the task tab. what do you think?
grafik


ivy-lli (approved):
Fine for me 👍

PR #567, opened 2024-04-23
XIVY-13806: Add improvement from feedback

ivy-edp:
I discussed the following points again with bruno and have now adjusted them:
- Output Data changed back to Output, as it is better to have only one word
- In the mail: Error section moved to its own (new) Error tab (analogous to the other elements, e.g. Rest...)
- Task-Switch-Gateway: Output-Tab was moved to the end like the other tabs


ivy-lli (approved):
👍

PR #582, opened 2024-05-15
Max code editor in script cell

ivy-edp:
I have added an additional expand button to all CodeEditorCells within the monaco editor as well as an expand icon to the condition tab of the database element in the header.

The monaco can now be opened in fullscreen for both.

beispiel_Expand


ivy-edp:
@ivy-lli Do you know, why this build has all of a sudden a problem with the inscription-core package?
grafik

ivy-lli (approved):
Still don't like this (max code editor for cells), but ok...
I think you need to adjust the setupTest file with new mocks, so that no monaco stuff leaks into the test run...

ivy-edp:
@ivy-lli The tests are now working again. I hope it is ok if I added the MaximizedCodeEditorBrowser to the index-file of the code-editors. What do you think?

PR #583, opened 2024-05-16
Fix browser styles

ivy-lli:
- Browser background color
- Tab border only on keyboard focus


ivy-lli:
I saw that in the design from anais is also the help text in a collapsible, but I think to improve also this design is a little bit to much for a quick adjustment.

ivy-edp (approved):

PR #586, opened 2024-05-21
XIVY-14079: Fix validation for maximized code-editor

ivy-edp:
In order for the maximized code editor (which I added to the expression cells and the condition in the db element), to display the correct validations in monaco, I had to forward the macro and type attributes.


ivy-cst (approved):
Looks good :+1:

PR #592, opened 2024-05-29
update-rpc-schema

ivy-lgi:


packages/protocol/schemaCodegen.cjs
```diff
@@ -17,7 +17,7 @@ tsGen.DEFAULT_OPTIONS.bannerComment = `
 `;
 
 const tsOut = path.resolve('./src/data/inscription.ts');
-var schemaUri = 'https://json-schema.ivyteam.ch/process/11.2.2/inscription.json';
+var schemaUri = 'https://json-schema.ivyteam.ch/process/11.4.0/inscription.json';
```
ivy-lgi, line 20: Can we automate this easily?

ivy-lgi: Maybe even the whole process of generating the protocol?

ivy-lli: Isn't this only the default schema uri, if no parameter is given? so most of the time should not make any difference?

And yes we maybe can automatically update the protocol and update create a PR, but I would say most of the time you will need to do manually changes too... so I'm not sure if it would be worth the effort...

ivy-lli (approved):

ivy-lli:
I will merge to fix the pipeline :)

packages/protocol/schemaCodegen.cjs
ivy-rew: yes, most of the time you will willingly create new features in the backend and then roll-out these types with a custom run. so the default uri doesn't matter much. 
in the backend we have now simplified the URIs... we will only generate one number matching the release. so we have to update once per release.
I wouldn't update/raise it automatically yet however, or we will have failing builds until the master has published the new version of the protocol.

ivy-rew:
:+1:

PR #594, opened 2024-06-04
XIVY-14204 Fix: NPE for Database-READ TableSort

ivy-edp:
- As with TableFields, an NPE occurred when switching from ANY to READ because orderBy was undefined.
- In addition, the defaultOpen was corrected accordingly.
- I wasn't quite sure if there was a shorter way of writing but I don't think so for these situations.

### DIFF
--- a/packages/editor/src/components/parts/query/database/TableSort.tsx
+++ b/packages/editor/src/components/parts/query/database/TableSort.tsx
@@ -43,7 +43,7 @@ export const TableSort = () => {
   const orderItems = useMemo<SelectItem[]>(() => Object.entries(QUERY_ORDER).map(([label, value]) => ({ label, value })), []);
 
   useEffect(() => {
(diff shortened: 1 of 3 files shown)


packages/editor/src/components/parts/query/database/TableSort.tsx
```diff
@@ -43,15 +43,17 @@ export const TableSort = () => {
...
+          const parts = order.split(' ');
+          const name = parts[0];
+          let sorting: OrderDirection = 'ASCENDING';
+          if (parts.length > 1 && parts[1] === 'DESC') {
+            sorting = 'DESCENDING';
+          }
+          return { name, sorting };
+        })
+      : [];
     setData(data);
```
ivy-lli, line 57: I would write it 
```ts
const data = config.query.sql.orderBy?.map...
setData(data ?? []);
```

ivy-lli (approved):

PR #596, opened 2024-06-11
Fix: Spacing between card-text-line with auto-word-wrap

ivy-edp:
The issue was that our custom text splitting into multiple CardLines maintained a 20px gap between them. However, if a span within a CardLine contained text longer than the card text div, it wrapped onto a new line, and this extra line wasn't considered when calculating the spacing for the subsequent CardLines.

I've now set the line-height to initial. If you prefer a larger gap between lines, you can set it to 20px, for instance.

Before:
grafik

After:
grafik

### DIFF
--- a/packages/editor/src/components/widgets/output/CardText.css
+++ b/packages/editor/src/components/widgets/output/CardText.css
@@ -12,8 +12,12 @@
   overflow: hidden;
   min-height: 14px;
 }
+
+.card-text > .card-text-line {
+  line-height: 25px;
+}
+
 .card-text .card-text-line {
-  height: 20px;
   white-space: pre-wrap;
 }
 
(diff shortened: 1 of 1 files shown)


packages/editor/src/components/widgets/output/CardText.css
```diff
@@ -12,8 +12,12 @@
...
 }
+
+.card-text > .card-text-line {
+  line-height: initial;
+}
+
 .card-text .card-text-line {
-  height: 20px;
   white-space: pre-wrap;
 }
```
ivy-lli, line 22: why new css block? can you just add this lineheight to the existing block?

ivy-lli (approved):
Does this work if a macro card is shown?
Can you merge this to 11.3.1 too?

ivy-edp:
CardArea:
grafik

CardText (singleline):
grafik

ivy-lli (approved):
Looks good thanks :)

PR #608, opened 2024-07-04
XIVY-14428 Improve Combobox design and responsivness

ivy-edp:
I have now adapted the combobox accordingly. Additional information is now displayed greyed out behind the "Main-Part". However, if there is not enough space for the additional information, it appears on the second line again and the words are wrapped accordingly so that you never have to scroll to the right to see the entire content.
responsiveCombobox

I deliberately chose not to scroll to the right here, as I have the feeling that scrolling in the open combobox may not be quite so pleasant. The only downside to this approach could be that if there is a lot of text in the info, you may again have the problem that the combobox dropdown becomes extremely long. What do you think? Is it more important to have everything on one line (a short dropdown) or see everything at once (multiple lines)?

### DIFF
--- a/packages/editor/src/components/parts/call/CallSelect.tsx
+++ b/packages/editor/src/components/parts/call/CallSelect.tsx
@@ -39,11 +39,9 @@ const CallSelect = ({ start, onChange, starts, startIcon }: CallSelectProps) => 
       <>
         <div>
           <IvyIcon icon={startIcon} />
-          <span style={item.deprecated ? { textDecoration: 'line-through' } : {}}>{item.startName}</span>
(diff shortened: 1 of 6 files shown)


ivy-lli (approved):
Nomally there should not be much content, but I think it is a good idea to wrap it. 
Let's see what bruno means to this solution.
Or maybe @ivy-rew has also a strong opinion here?

ivy-rew:
Looks much better thanks.  

Though there still will be often shown multiple lines ... since in the projects customers tend to have large namespaces :) 
So maybe abbreviating the text and only showing the full text on hover could also work. :thinking: 
But no strong preference here on my side :wink:

PR #610, opened 2024-07-04
XIVY-14428 Further improve Combobox-Layout

ivy-edp:
I played around a bit with the combobox layout and added a tooltip option to the combobox. 
@ivy-lli, @ivy-rew  What do you think of this solution? I find it quite elegant. The only disadvantage is that there is no tooltip on mobile devices.
furtherImproveDialog

@ivy-lli Initially I wanted to add the tooltip directly in the comboboxItem within the Call itself, but unfortunately I always got a React error message here: "React.Children.only expected to receive a single React element child."
I have now simply extended the comboboxItem with the tooltip, but have to work with "as T" at the end in certain places to get rid of TypeScript error messages. What do you think? Maybe this tooltip complicates the combobox too much. 

I will certainly wait with this merge until we have also received feedback from bruno.

### DIFF
--- a/packages/editor/src/components/parts/call/CallSelect.tsx
+++ b/packages/editor/src/components/parts/call/CallSelect.tsx
@@ -17,7 +17,11 @@ const CallSelect = ({ start, onChange, starts, startIcon }: CallSelectProps) => 
   const items = useMemo<CallableStartItem[]>(
     () =>
       starts.map(start => {
-        return { ...start, value: start.id };
+        return {
+          ...start,
+          value: start.id,
+          tooltip: { main: `${start.process} : ${start.startName}`, additional: `${start.project} ${start.packageName}` }
+        };
       }),
     [starts]
   );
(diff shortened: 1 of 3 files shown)


ivy-lli (commented):
Definilty possible, IMO a little bit overengeeniert...
A simple title on each option which contains the value would be totally fine for me. This would also provide a little "tooltip" on hover. 
Do we really now that this will be a problem? Maybe most of the people have this inscription view much wider and they don't have this problem anyways...
At best also with a aria-label so the accessibility is better :)

ivy-lli:
> Oh yes, you're right, that would effectively have been much easier 🤦‍♂️. If we decide to solve the whole thing with a tooltip without a wrap, I will add title and aria-label instead of working with the tooltip component. grafik

I mean not that the tooltip is a bad component. I have a little bit of respect on adding to much stuff in places where a loop creates many of such elements.

ivy-rew:
From a user perspective, the tooltip solution looks great to me. I think dialogs namespaces and parameteres will often use a lot of space. So I think rarely will the screen be wide enough to show everything without a tooltip.

ivy-rew:
If easily possible, we could also think of 'omiting' all namespaces within the combo for 'parameter' and 'return types' ... this would also free up a lot of space, in the workspace of the GIFs you have recorded.

ivy-edp:
@ivy-rew Not a bad idea. I can also discuss it with Bruno. However, I think for the moment it's okay if we simply arrange the dialogue combo box like this: MainDlg : start(data) - package name

I have now made a new PR, in which I have worked with title and aria-label and changed everything to single-line. I think this is fine for now. I'll close this PR, but leave the branch open in case we still prefer the tooltip with the tooltip component.

PR #639, opened 2024-10-09
Stabilize test

ivy-lli:


packages/editor/src/components/parts/common/table/useResizableEditableTable.tsx
```diff
@@ -61,12 +61,10 @@ const useResizableEditableTable = <TData,>({
   useEffect(() => {
```
ivy-lli, line 61: In general I think this was the real problem.
However I hadn't the currage to change this to something else and break some UI, so I fixed it in the MailAttachmentPart directly.

packages/editor/src/components/parts/mail/MailAttachmentPart.tsx
```diff
@@ -46,16 +48,15 @@ const MailAttachmentTable = () => {
   );
 
   const onChange = (change: MailAttachment[]) => {
-    setData(change);
```
ivy-lli, line 49: This is mostly the fix. I changed it from an internal state to only a memo value.

The problem was when there was an empty row on the table and after a reload the useEffect tried to remove it. Then it triggers a onChange here which changes the data which changes the table which triggers the useEffect again. Or at least something like that.
With the memo we always depend on the global state and only modify it to our needs in here.

ivy-edp (commented):
thanks 👍

PR #640, opened 2024-10-10
XIVY-15045 Fix renaming of DataClass to DataclassType

ivy-edp:
Will result due to this change: https://github.com/axonivy/core/pull/6964

### DIFF
--- a/packages/protocol/src/data/inscription.ts
+++ b/packages/protocol/src/data/inscription.ts
@@ -38,7 +38,7 @@ export interface Inscription {
   databaseColumn: DatabaseColumn[];
   databaseColumnRequest: DatabaseColumnRequest;
   databaseTablesRequest: DatabaseTablesRequest;
-  dataClass: DataClass[];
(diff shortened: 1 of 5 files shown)


packages/protocol/src/data/inscription.ts
```diff
@@ -721,10 +720,6 @@ export interface OutlineNode {
   title: string;
   type: string;
 }
-export interface ProcessMetaRequest {
-  context: InscriptionContext;
-  processId: string;
-}
```
ivy-lli, line 727: not used?

ivy-lli: I think this one is from the outline...

ivy-edp: Possibly, but I don't know for sure. I simply generated the latest inscription.ts from my core branch and replaced it.

ivy-lli (approved):
Seems no longer in usage (just uses the InscriptionElementContext)

PR #641, opened 2024-10-14
Use npm instead of yarn

ivy-lli:


ivy-lli:
fyi @ivy-cst @ivy-edp

PR #648, opened 2024-10-24
XIVY-14728 replaces CardText with InputBadge

ivy-fhe:


packages/editor/src/components/widgets/code-editor/ScriptInput.css
```diff
@@ -16,3 +16,7 @@
   right: 0;
   margin: var(--size-2);
 }
+
+.input-badge-wrapper {
+  width: 100%;
+}
```
ivy-lli, line 22: why needed?

ivy-fhe: As it's not set on the component, it's needed that the input uses the full width
if it's set on the component, it would lead to problems in the form-editor

ivy-fhe: Then it just runs out the side :smile: 
image

ivy-lli: if you do it the same as it is on the input with calc(100% - 2 * paddings)?

ivy-lli (approved):

PR #650, opened 2024-11-06
XIVY-14823 Add NewRole-Popover to RoleBrowser

ivy-edp:
My proposal for the AddRole functionality in the Type Browser (Core: https://github.com/axonivy/core/pull/7186). 

Certain validations are currently implemented directly in the frontend; it might make more sense to move these to the backend in the future?

Open To-Dos:
- Write tests
- Equip the CMS browser with the same query invalidation for the Add CMS functionality
- Later, I might try to have the newly added role automatically selected in a separate PR. What do you think of this feature?
newRole

### DIFF
--- a/packages/editor/src/components/browser/Browser.tsx
+++ b/packages/editor/src/components/browser/Browser.tsx
@@ -55,7 +55,7 @@ const Browser = ({ open, onOpenChange, types, accept, location, cmsOptions, role
 
   return (
     <>
-      <Dialog open={open} onOpenChange={onOpenChange}>
+      <Dialog open={open} onOpenChange={onOpenChange} modal={false}>
         <DialogTrigger asChild>
           <Button icon={IvyIcons.ListSearch} aria-label='Browser' />
         </DialogTrigger>
--- a/packages/editor/src/components/browser/role/AddRolePopover.tsx
+++ b/packages/editor/src/components/browser/role/AddRolePopover.tsx
@@ -1,0 +1,90 @@
+import { useState } from 'react';
+import {
+  BasicField,
+  Button,
+  Flex,
+  Input,
+  Message,
+  Popover,
+  PopoverArrow,
+  PopoverContent,
+  PopoverTrigger,
+  toast
+} from '@axonivy/ui-components';
+import { IvyIcons } from '@axonivy/ui-icons';
+import { useFunction } from '../../../context/useFunction';
+import { useQueryClient } from '@tanstack/react-query';
+import { useEditorContext } from '../../../context';
+import type { RoleMeta } from '@axonivy/inscription-protocol';
+import type { Table } from '@tanstack/react-table';
+import { useRoles } from '../../parts/common/responsible/useRoles';
+import { isValidRowSelected, newNameExists, newNameIsValid } from './validate-role';
+
+export const AddRolePopover = ({ value, table }: { value: string; table: Table<RoleMeta> }) => {
+  const [open, setOpen] = useState(false);
+  const { taskRoles } = useRoles();
+  const [newRoleName, setNewRoleName] = useState('');
+
+  const queryClient = useQueryClient();
+  const { context } = useEditorContext();
+  const addRole = useFunction(
+    'meta/workflow/addRole',
+    {
+      context,
+      newRole: { identifier: 'test', parent: 'Everybody' }
+    },
+    {
+      onSuccess: () => {
+        toast.info('Role successfully added');
+        queryClient.invalidateQueries({ queryKey: ['meta/workflow/roleTree', context] });
+        setOpen(false);
+      },
+      onError: error => {
+        toast.error('Failed to add new role', { description: error.message });
+      }
+    }
+  );
+
+  return (
+    <Popover
+      open={open}
+      onOpenChange={e => {
+        setOpen(e);
+        if (e) setNewRoleName('');
+      }}
+    >
+      <PopoverTrigger asChild>
+        <Button
+          icon={IvyIcons.Plus}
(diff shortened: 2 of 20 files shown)


packages/editor/src/components/browser/Browser.tsx
```diff
@@ -55,7 +55,7 @@ const Browser = ({ open, onOpenChange, types, accept, location, cmsOptions, role
 
   return (
     <>
-      <Dialog open={open} onOpenChange={onOpenChange}>
+      <Dialog open={open} onOpenChange={onOpenChange} modal={false}>
```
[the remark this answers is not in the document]
ivy-lli: ahhh currently not possible via the ui-components stuff... there all is attached to the body.
Btw, the problem in this case it the inscription dialog which is not attached to the body 🙈

packages/editor/src/components/browser/role/AddRolePopover.tsx
```diff
@@ -0,0 +1,110 @@
...
+        <Button icon={IvyIcons.Plus} aria-label={`Add Role to ${value}`} title={`Add Role to ${value}`} disabled={!isValidRowSelected()} />
+      </PopoverTrigger>
+      <PopoverContent sideOffset={12} collisionPadding={5} style={{ zIndex: '10000' }}>
+        <Flex direction='column' gap={2} alignItems='center'>
+          <BasicField label='New role name' style={{ width: '100%' }}>
+            <Input value={newRoleName} onChange={e => setNewRoleName(e.target.value)} style={{ zIndex: '1000' }} />
+          </BasicField>
+          {newNameExists() && <Message variant='error' message='A role with that name already exists' />}
+          <Button
+            key='addRole'
```
ivy-lli, line 90: why is this needed?

packages/editor/src/components/browser/role/AddRolePopover.tsx
```diff
@@ -0,0 +1,110 @@
...
+        if (e) setNewRoleName('');
+      }}
+    >
+      <PopoverTrigger asChild>
+        <Button icon={IvyIcons.Plus} aria-label={`Add Role to ${value}`} title={`Add Role to ${value}`} disabled={!isValidRowSelected()} />
+      </PopoverTrigger>
+      <PopoverContent sideOffset={12} collisionPadding={5} style={{ zIndex: '10000' }}>
+        <Flex direction='column' gap={2} alignItems='center'>
+          <BasicField label='New role name' style={{ width: '100%' }}>
+            <Input value={newRoleName} onChange={e => setNewRoleName(e.target.value)} style={{ zIndex: '1000' }} />
```
ivy-lli, line 86: why is this z-index needed? I understand the one on the content, but not sure about this one

packages/editor/src/components/browser/role/AddRolePopover.tsx
```diff
@@ -0,0 +1,110 @@
...
+        setOpen(e);
+        if (e) setNewRoleName('');
+      }}
+    >
+      <PopoverTrigger asChild>
+        <Button icon={IvyIcons.Plus} aria-label={`Add Role to ${value}`} title={`Add Role to ${value}`} disabled={!isValidRowSelected()} />
+      </PopoverTrigger>
+      <PopoverContent sideOffset={12} collisionPadding={5} style={{ zIndex: '10000' }}>
+        <Flex direction='column' gap={2} alignItems='center'>
+          <BasicField label='New role name' style={{ width: '100%' }}>
```
ivy-lli, line 85: why are the width 100% needed here on the field and on the button? I think this should already work when the parent flex is in column mode...

ivy-lli (approved):
Looks good in general. Some suggestions, questions below.
By the way, you could easily test your validation functions with unit tests if you would move them out of the component.

I think it would be cool if the new created role would be selected afterwards 👍

packages/editor/src/components/browser/role/AddRolePopover.tsx
ivy-edp: Unfortunately not, if I remove the width 100% and only the flex is on column (even if I set the parentFlex width to 100%, it looks like this:
grafik

However, I would like it to look like this:
grafik

packages/editor/src/components/browser/role/AddRolePopover.tsx
ivy-edp: Yes, you are right. This is not needed 👍

packages/editor/src/components/browser/Browser.tsx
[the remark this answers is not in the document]
ivy-edp: Yes exactly, had to change this to false, otherwise you couldn't put the focus on the popover.
All right, yes maybe we could improve the UI component in the future so that you can define the container of the popover.

playwright/tests/integration/mock/browser/tableCol.spec.ts
```diff
@@ -56,19 +56,20 @@ test('browser add table column doubleclick', async ({ page }) => {
 
 async function applyTableColBrowser(page: Page, expectedSelection: string = '', rowToCheck: number, numberOfRows?: number, dblClick?: boolean) {
   await browserBtn(page).nth(0).click();
-  await expect(page.getByRole('dialog')).toBeVisible();
-  await page.getByText('Table Column').first().click();
+  const browserDialog = page.getByRole('dialog');
```
ivy-edp, line 59: @ivy-lli I believe the only side effect I could identify from setting modal=false in our tests is that I now had to specifically target the dialog, as page.getByRole('row') was also picking up rows from behind the browser dialog. However, I don't think this is particularly concerning, do you?

ivy-lli (approved):

PR #651, opened 2024-11-07
XIVY-14823 refactor newCmsString-Action to a function

ivy-edp:
I would like to remove the refresh button from the CMS browser so that whenever a new CMS string is added via the SWT dialog, the data is refreshed automatically. I am using the same architecture as in the combine function in the data class editor for this.

newCmsString

### DIFF
--- a/packages/editor/src/components/browser/cms/CmsBrowser.tsx
+++ b/packages/editor/src/components/browser/cms/CmsBrowser.tsx
@@ -1,13 +1,15 @@
 import { useMemo, useEffect, useState } from 'react';
-import { ActionCell, Checkbox, ExpandableCell, SearchTable } from '../../widgets';
+import { Checkbox, ExpandableCell, SearchTable } from '../../widgets';
 import type { UseBrowserImplReturnValue } from '../useBrowser';
-import { useAction, useEditorContext, useMeta } from '../../../context';
+import { useEditorContext, useMeta } from '../../../context';
 import type { ColumnDef, ColumnFiltersState, ExpandedState, RowSelectionState, VisibilityState } from '@tanstack/react-table';
 import { flexRender, getCoreRowModel, getExpandedRowModel, getFilteredRowModel, useReactTable } from '@tanstack/react-table';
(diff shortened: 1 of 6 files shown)


ivy-edp:
Unfortunately, I still have a problem at the moment. If you add several strings directly, only the first string is refreshed with onSuccess. I assume that the core does not wait until all new CMS strings have been entered. Don't quiet now how to solve this. Maybe @ivy-rew has an idea.

newCmsStringMulti

ivy-edp:
I will also try to change the design (so that it is the same as the Role Browser --> Plus top right) in a later PR. It's a bit more difficult, though, because you can't currently select the folders in the CMS browser, because it makes no sense to add a folder via the browser. But a possible solution would be to deactivate Apply if a folder is selected. @ivy-lli What do you think?

packages/editor/src/components/browser/cms/CmsBrowser.tsx
```diff
@@ -155,10 +173,7 @@ const CmsBrowser = ({ value, onChange, noApiCall, typeFilter, onDoubleClick, loc
 
   return (
     <>
-      <div className='browser-table-header'>
-        <Checkbox label='Enable required Projects' value={requiredProject} onChange={() => setRequiredProject(!requiredProject)} />
-        <Button onClick={() => refetch()} title='Refresh CMS-Browser' aria-label='refresh' icon={IvyIcons.Redo} />
-      </div>
+      <Checkbox label='Enable required Projects' value={requiredProject} onChange={() => setRequiredProject(!requiredProject)} />
```
ivy-lli, line 176: what happens if you select a node from a different project as parent for the new cms entry? 
I think we need this new cms thing only on the own project, so we should then disable the button for others

ivy-lli (approved):
Yes I would move the add button to the top, and disable the apply button if a folder is selected 👍

ivy-edp:
It looks now like this:
grafik

ivy-lli (approved):

PR #653, opened 2024-11-08
XIVY-14823 Autoselect and expand newly added role

ivy-edp:
To implement this auto-select feature, I had to use two useEffect hooks. Initially, I thought I could just change the selection in onSuccess, but it takes some time for the table to render with the new data. I could either use a setTimeout in onSuccess (not ideal) or use these useEffects to update the selection when a new role is detected. What do you think?

autoSelectNewRole

### DIFF
--- a/packages/editor/src/components/browser/role/RoleBrowser.tsx
+++ b/packages/editor/src/components/browser/role/RoleBrowser.tsx
@@ -92,10 +92,23 @@ const RoleBrowser = (props: {
     props.onChange({ cursorValue: selectedRow.original.id });
   }, [props, rowSelection, table]);
 
+  const [addedRole, setAddedRoleName] = useState('');
+
+  useEffect(() => {
+    if (addedRole.length === 0) {
+      return;
+    }
+    const newRow = table.getRowModel().flatRows.find(row => row.original.id === addedRole);
+    if (newRow) {
+      newRow.getParentRow()?.toggleExpanded(true);
+      setRowSelection({ [newRow.id]: true });
+    }
+  }, [addedRole, roleItems, table]);
+
   return (
     <>
       <Flex justifyContent='flex-end'>
-        <AddRolePopover value={props.value} table={table} />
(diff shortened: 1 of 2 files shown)


ivy-edp:
Alternatively, I could solve it with a setTimeout directly in addRole --> onSuccess (without using useEffect). However, I don't think this approach is much better:
grafik

ivy-lli (commented):
I'm definitly against the setTimeout
But do you want compare the role trees between different fetches to check for new roles?

Didn't you have all the information from the user input? The parent role you expect this new role to appear as the role name? 
So all you have to do is to save this somewhere, and when data comes and you have such saved data use it to select the corresponding row.

packages/editor/src/components/browser/role/RoleBrowser.tsx
```diff
@@ -92,10 +92,26 @@ const RoleBrowser = (props: {
...
+  useEffect(() => {
+    const newRow = table.getRowModel().flatRows.find(row => row.original.id === addedRole);
+    if (newRow) {
+      const parentRows = newRow.getParentRows();
+      const expandedState = parentRows.reduce((acc, parentRow) => {
+        acc[parentRow.id] = true;
+        return acc;
+      }, {} as Record<string, boolean>);
+
+      setExpanded(expandedState);
```
ivy-lli, line 106: Techically the parent is already selected, so at least to it everything is expanded right?
So then isn't it enough to make sure this row is expanded?

```suggestion
      newRow.getParentRow().toggleExpanded(true);
```

ivy-lli (commented):
Yes exaclty. This is much simpler or not?

packages/editor/src/components/browser/role/RoleBrowser.tsx
ivy-edp: Yes, you're right, thank you. It works the same way and is much easier 👍

ivy-lli (approved):

ivy-lli:
Just something poped into my mind.
I think it would be a good idea to reset the addedRoleName state to '' after change the selection.

PR #656, opened 2024-11-11
XIVY-14823 Reset newRoleName on autoSelection & add role via enter

ivy-edp:
As you suggested, I reset the addedRoleName after the selection.
I have also changed the appearance of the Add Role button to ‘Primary’ and added the option to add the role directly with ‘Enter’ after you typed in a valid role name.
grafik

### DIFF
--- a/packages/editor/src/components/browser/role/AddRolePopover.tsx
+++ b/packages/editor/src/components/browser/role/AddRolePopover.tsx
@@ -69,29 +69,33 @@ export const AddRolePopover = ({
           disabled={!isValidRowSelected(table, taskRoles)}
         />
       </PopoverTrigger>
-      <PopoverContent sideOffset={12} collisionPadding={5} style={{ zIndex: '10000' }}>
-        <Flex direction='column' gap={2} alignItems='center'>
-          <BasicField label='New role name' style={{ width: '100%' }}>
-            <Input value={newRoleName} onChange={e => setNewRoleName(e.target.value)} />
-          </BasicField>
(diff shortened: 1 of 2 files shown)


packages/editor/src/components/browser/role/AddRolePopover.tsx
```diff
@@ -70,28 +70,31 @@ export const AddRolePopover = ({
...
+                  context,
+                  newRole: { identifier: newRoleName, parent: value }
+                });
+                setAddedRoleName(newRoleName);
+              }}
+              aria-label='Add new Role'
+              title='Add new Role'
+              disabled={!newNameIsValid(table, newRoleName)}
+              style={{ width: '100%' }}
+              variant='primary'
```
ivy-lli, line 92: if you use a from with onSubmit then I think you should add the `type='submit'` to the button.

packages/editor/src/components/browser/role/RoleBrowser.tsx
```diff
@@ -102,6 +102,7 @@ const RoleBrowser = (props: {
     if (newRow) {
       newRow.getParentRow()?.toggleExpanded(true);
       setRowSelection({ [newRow.id]: true });
+      setAddedRoleName('');
```
ivy-lli, line 105: thx

ivy-lli (approved):

ivy-lli:
Just saw it in the screenshot, I think the popover is a little bit far away from the add button or not?

PR #662, opened 2024-11-14
XIVY-15201 include 'help' section in browser screens :camera_flash: 

ivy-rew:


playwright/inscription-test-project/dataclasses/ch/ivyteam/test/Person.d.json
```diff
@@ -2,6 +2,7 @@
   "$schema" : "https://json-schema.axonivy.com/data-class/12.0.0/data-class.json",
   "simpleName" : "Person",
   "namespace" : "ch.ivyteam.test",
+  "comment" : "A registrated webshop user",
```
ivy-rew: tried to show this in the help; but it seems like we don't send the 'description comment' as help for dataclasses :shrug:

playwright/tests/screenshots/browsers.spec.ts
```diff
@@ -17,22 +17,25 @@ test.describe('Browsers', () => {
...
     await browser.search('role');
-    await expect(browser.table.getByRole('row', { name: 'ivy' })).toBeVisible();
+    await browser.table.getByRole('row', { name: 'security :' }).click();
+    await expect(browser.help).toContainText('ivy.security');
     await screenshot(dialog, 'browser-function.png');
 
     await browser.openTab('Type');
     await browser.search('Per');
-    await expect(browser.table.getByRole('cell', { name: 'Person :' })).toBeVisible();
+    await browser.table.getByRole('row', { name: 'Person :' }).first().click();
```
ivy-rew, line 38: couldn't display the "help" here in a predictable manor. 
first, there is no real addition ... the help just shows the select value ... here "Person". I would have expected it to at least show the DataClass description. Not very important but seems an easy one to improve on the backend.
second, asserting it to be present works on my machine ... not on the CI.

ivy-edp: Yes, it is the case that the help doesn't really help much with the typebrowse and only shows the name of the type for the moment. This could of course be improved. I think for the moment, a screenshot of the help only makes sense for the CMS & Functionbrowser.

Unfortunately I don't see why the assertion doesn't work. Maybe @ivy-lli has an idea.

ivy-edp (commented):
Overall, it looks good for me👍. However, I’m not sure if all browsers need the screenshot of the help, as it currently only really assists with the CMS and Function Browser.

playwright/tests/screenshots/browsers.spec.ts
ivy-rew: > Unfortunately I don't see why the assertion doesn't work. Maybe @ivy-lli has an idea.

Not really important for me now; I see that the clicked column becomes kind of un-selected after clicking it. Could surely be solved by re-clicking. But I think it could be a pointer to a real problem that exits... we may stumble over this again when streamlining the browsers :shrug:

ivy-rew:
> Overall, it looks good for me👍. However, I’m not sure if all browsers need the screenshot of the help, as it currently only really assists with the CMS and Function Browser.

I'd show the help also for the other browsers while it exists. E.g. roles would also have a description that could be displayed. Now it's apparent that it isn't displayed onto the UI ... but we can improve and quickly see the change manifested in screenshots once we address it.

ivy-lli (approved):

PR #665, opened 2024-11-21
XIVY-15332 Use normal table if rows are not expanded in function browser

ivy-edp:
Due to setExpanded, the initial state can only be set to true, otherwise the rows won't render (the exact reason is unclear). Initially, I use the regular table, and once any "ivy" children are expanded (either manually or via search), the virtual table is used.

Also, I refactored the sortedTree useEffect into a useMemo. I think this is better for re-rendering purposes, right?
example_function

### DIFF
--- a/packages/editor/src/components/browser/function/FunctionBrowser.tsx
+++ b/packages/editor/src/components/browser/function/FunctionBrowser.tsx
@@ -9,15 +9,15 @@ import {
   type RowSelectionState,
   getCoreRowModel,
   getExpandedRowModel,
-  getFilteredRowModel,
-  flexRender
+  getFilteredRowModel
 } from '@tanstack/react-table';
 import { useVirtualizer } from '@tanstack/react-virtual';
 import { IvyIcons } from '@axonivy/ui-icons';
 import type { BrowserValue } from '../Browser';
 import { useEditorContext, useMeta } from '../../../context';
 import { getParentNames } from './parent-name';
-import { SelectRow, TableBody, TableCell, TableFooter } from '@axonivy/ui-components';
+import { TableBody, TableFooter } from '@axonivy/ui-components';
+import BrowserTableRow from '../BrowserTableRow';
 export const FUNCTION_BROWSER_ID = 'func' as const;
 
 export const useFuncBrowser = (onDoubleClick: () => void): UseBrowserImplReturnValue => {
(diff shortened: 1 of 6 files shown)


packages/editor/src/components/browser/function/FunctionBrowser.tsx
```diff
@@ -36,37 +36,32 @@ const FunctionBrowser = (props: { value: string; onChange: (value: BrowserValue)
   const [method, setMethod] = useState('');
   const [paramTypes, setParamTypes] = useState<string[]>([]);
   const [type, setType] = useState('');
-  const [sortedTree, setSortedTree] = useState<Function[]>([]);
   const { data: tree } = useMeta('meta/scripting/functions', undefined, []);
   const { data: doc } = useMeta('meta/scripting/apiDoc', { context, method, paramTypes, type }, '');
+  const sortedTree = useMemo(() => {
+    if (!tree || tree.length === 0) {
+      return [];
```
ivy-lli, line 43: have an initial tree here doesn't work with the init expanded state + virtualize?

packages/editor/src/components/browser/function/FunctionBrowser.tsx
```diff
@@ -166,16 +161,24 @@ const FunctionBrowser = (props: { value: string; onChange: (value: BrowserValue)
...
-          })}
+          {rows.some(row => row.depth === 1 && row.getIsExpanded())
+            ? virtualizer.getVirtualItems().map(virtualRow => {
+                const row = rows[virtualRow.index];
+                return (
+                  <SelectRow key={row.id} row={row} onDoubleClick={props.onDoubleClick}>
+                    {row.getVisibleCells().map(cell => (
+                      <TableCell key={cell.id}>{flexRender(cell.column.columnDef.cell, cell.getContext())}</TableCell>
+                    ))}
+                  </SelectRow>
```
ivy-lli: duplicated now. maybe strip out shared component for this?

packages/editor/src/components/browser/function/FunctionBrowser.tsx
ivy-edp: No, unfortunately not 😬

ivy-lli (approved):
I would deduplicate the SelectedRow, but then I think good for merge :)
Thanks for the improvement with the useMemo 👍

packages/editor/src/components/browser/function/FunctionBrowser.tsx
ivy-edp: Sure, I can do that. Should I implement this new Row component at the browser level, as I could reuse it for the Role, Type, and Attribute Browsers as well? However, this will likely become unnecessary soon, as we plan to use the UI component browser soon.
grafik

ivy-lli: if they are the same why not 👍

ivy-lli (approved):

PR #666, opened 2024-11-22
XIVY-15332 Only enable apiDoc-Meta for Pro Designer

ivy-edp:
The JdtApiDocService does not work on neo (ivy.log gets flooded with warnings). I disabled the useQuery for the apiDoc if the app is not "designer".


ivy-lli:
Can't you disable this endpoint in the backend? 
Maybe that the endpoint returns a message 'JavaDoc is currently not supported on the engine'

ivy-edp:
I think I could do it that way. I took inspiration from the dataclass editor, where we also disable metas on the frontend using 'enable.' However, I think this isn't quite the same scenario. So should I instead fix this on the backend?

ivy-lli:
I think disable a button should be done in the frontend. but this is show info for a selection. this should be handled by the backend...

ivy-lli:
The process inscription code now lives in the [process-editor-client](https://github.com/axonivy/process-editor-client) repo.
This repo get archived.

